PR #13, opened 2015-03-27
Update server.js

CooCooooo:
a solution to     bsonInstance = new BSON(bsonTypes);
                                               ^
                       TypeError: undefined is not a function


christkv:
what os and script to reproduce. the fix you propose it's not the right one.

CooCooooo:
I use the window8.1 and can't have the BSON compile, I work well with the change

christkv:
it's not broken the console message you get means it could not use the c++ parser and falls back to the javascript one.

CooCooooo:
yes, but if you can get it work for someone who don't have the c++ parser may be a good choice   ^_^

christkv:
bumped bson to 3.0.1, delete your node_modules directory and do npm install and tell me if it fixes the issue.

PR #15, opened 2015-04-09
Minor cleanups and package upgrades

prabhu:


christkv:
code has moved since then

PR #16, opened 2015-04-10
Updated default value of param options.rejectUnauthorized to match documentation

sunknudsen:
When connecting to a SSL MongoDB server configured to use a self-signed certificate using the node-mongodb-native node library, the connection fails with MongoError DEPTH_ZERO_SELF_SIGNED_CERT. I discovered the error is linked to mongodb-core. According to http://mongodb.github.io/node-mongodb-native/core/tutorials/connecting/, the default options.rejectUnauthorized value should be false which is mandatory when using self-signed certificates.

### DIFF
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -33,7 +33,7 @@ var debugFields = ['host', 'port', 'size', 'keepAlive', 'keepAliveInitialDelay',
  * @param {Buffer} [options.cert] SSL Certificate binary buffer
  * @param {Buffer} [options.key] SSL Key file binary buffer
  * @param {string} [options.passphrase] SSL Certificate pass phrase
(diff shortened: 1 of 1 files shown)


christkv:
actually the docs need updating it seems as they are supposed to be rejected by default. it was done on purpose to ensure safest setting as default and disable at your own risk.

sunknudsen:
Thanks for the update. How can this be configured at the node-mongodb-native library level without custom hacks in mongodb-core?

christkv:
http://mongodb.github.io/node-mongodb-native/2.0/tutorials/connecting/

sslValidate = false

christkv:
have a look at this

http://mongodb.github.io/node-mongodb-native/2.0/tutorials/enterprise_features/

PR #27, opened 2015-06-20
handle writeErrors

dsuket:
throw errors when result has writeErrors.


christkv:
fixed in https://jira.mongodb.org/browse/NODE-526

PR #44, opened 2015-08-20
Still allow primary only connection if secondaryOnly is true

fastest963:
If you pass `connectWithNoPrimary` then `secondaryOnlyConnectionAllowed` is set to true but then this breaks a primary-only connection state. If `secondaryOnlyConnectionAllowed` is true then `isConnected` just returned if there was a secondary connected instead of allowing just a primary to still be connected.


fastest963:
Bug: https://jira.mongodb.org/browse/NODE-535

PR #64, opened 2015-12-09
use process.nextTick for handleCallback

iamruinous:
Prevent Maximum call stack exceeded by calling all callbacks on nextTick


christkv:
backed this out as it break node 0.10.x and earlier support. can you provide a code example that causes the issue as no-one else seems to have run into this.

PR #70, opened 2015-12-28
Allow connection to secondary if primaryPreferred or secondaryPreferred

leichter:
If the read preference is `primaryPreferred` or `secondaryPreferred`, `isConnected()` should return true if either a primary or secondary node are connected.

JIRA ticket: [https://jira.mongodb.org/browse/NODE-628](https://jira.mongodb.org/browse/NODE-628)


christkv:
can you provide a test or test script that would cause this issue please.

leichter:
Ok, I'll see what I can do

leichter:
Added a test to the replica set read preference tests; this will fail when my additions in `lib/topologies/replset.js` are commented out.

PR #71, opened 2016-01-03
Initial implementation of a pool implemented using a queue and workers

owenallenaz:
This is the initial implementation for a pool that uses a queue worker as opposed to a round robin strategy. It works in a non-destructive fashion so that we didn't have to refactor the code base in entirety to handle the new pattern, rather we can gradually switch over component by component to the new mechanism until all commands are eventually migrated. This is meant for review, and then I'll circle back and correct issues or disagreements.

In example, in this current pull request I've utilized the strategy for initial cursor, getMore cursor, and Server.command.

Basically when a query comes in, it stashes the callback and the query in a queue and then executes to process the queue. If there is a connection available, it will have that connection process it, upon completion of another query, it will also process the queue.

Here is an example of the test script I used: https://gist.github.com/owenallenaz/f5a8497e977c45d013ff 

If you execute that script before my pull, you'll see many queries stall behind the slowFn. Using my pull request all of the queries will execute instantly and none of them will stall behind the slowFn, allowing for a much more efficient use of the connection pool.

Here are some outstanding issues that still need to be addressed:

1. I'm not familiar with the exhaust component of the cursor, I left that area untouched. How do I reproduce an exhaust cursor.

2. For the monitor, it seems to me the easiest way to maintain the HA monitor would be to have it's own connection pool of size 1, instead of having a reserved connection on the main pool. In this way we abstract away the complications of the monitor pool away from the operations of our main pool. What are your thoughts?

3. Is there a reason the monitor is implemented with a setTimeout in many different places in the code instead of a single setInterval? It seems like it would make a lot of that pathway much simpler.

4. Why is it that if a single net.connection in a pool errors or times out the whole pool dies? Shouldn't the pool regenerate one connection at a time, rather than killing the entirety of the pool off on a single connection error?

5. In cursor, execInitialQuery seems like dead code. Was it left behind for a reason, or could I remove?

6. What is the purpose of the onAll capability. I could not find reference to queries utilizing this. Is it entirely for auth?

All of the above issues aside, the only piece that prevents this pull request from being integrated is the monitoring component as it currently does not address that. So let me know your thoughts and I can finish that piece up.

### DIFF
--- a/lib/connection/pool.js
+++ b/lib/connection/pool.js
@@ -58,6 +58,7 @@ var Pool = function(options) {
   // Contains all connections
   this.connections = [];
   this.available = [];
+  this.monitorAvailable = [];
   this.state = DISCONNECTED;
   this.queryQueue = [];
   // Round robin index
@@ -127,8 +128,13 @@ var connectHandler = function(self) {
   return function(connection) {
     self.connections.push(connection);
     self.available.push(connection);
+    
     // We have connected to all servers
     if(self.connections.length == self.size) {
+      if (self.monitoring === true) {
+        self.monitorAvailable.push(self.connections[self.connections.length - 1]);
+      }
+      
       self.state = CONNECTED;
       // Done connecting
       self.emit("connect", self);
@@ -162,6 +168,7 @@ Pool.prototype.destroy = function(err) {
   
   this.queryQueue = [];
   this.available = [];
+  this.monitorAvailable = [];
 }
 
 var execute = null;
@@ -317,14 +324,19 @@ Pool.prototype._processQueue = function() {
   var self = this;
   
   // check to see if there is queries in queue and available connections to process
-  if (self.queryQueue.length === 0 || self.available.length === 0) { return; }
+  if (self.queryQueue.length === 0) { return; }
+  
+  var possibleQuery = self.queryQueue[0];
+  var connectionPool = possibleQuery.monitoring === true && self.monitoring === true ? self.monitorAvailable : self.available;
+  
+  if (connectionPool.length === 0) { return; }
   
   var queryArgs = self.queryQueue.shift();
-  var connection = self.available.shift();
+  var connection = connectionPool.shift();
   
   var complete = function(err, result) {
     // make our connection available again
-    self.available.push(connection);
+    connectionPool.push(connection);
     
     // pass the returned values to our original callback
     queryArgs.cb.apply(null, arguments);
--- a/lib/connection/pool.js
+++ b/lib/connection/pool.js
@@ -248,27 +248,27 @@ Pool.prototype.get = function(options) {
  * @param {number} maxConnections reduce the poolsize to maxConnections
  */
 Pool.prototype.capConnections = function(maxConnections) {
-    while(this.connections.length > maxConnections) {
-      // removes connections off the front so that our monitor socket (the last socket) is intact
-      var connection = this.connections.shift();
-      connection.removeAllListeners('close');
-      connection.removeAllListeners('error');
-      connection.removeAllListeners('timeout');
-      connection.removeAllListeners('parseError');
-      connection.removeAllListeners('connect');
-      connection.destroy();
-      
-      // if this connection is in the available pool, make sure it's popped off the queue
-      var index = this.available.indexOf(connection);
-      if (index > -1) {
-        this.available.splice(index, 1);
-      }
-    }
+  while(this.connections.length > maxConnections) {
+    // removes connections off the front so that our monitor socket (the last socket) is intact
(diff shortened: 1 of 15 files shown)


christkv:
1. Exhaust will go away so you can ignore it. It's a remanent of bygone times on MongoDB.
2. The design was picked to avoid having in effect two different sets of connections open at a given time and having to double handle authentication both in the monitor as well as the driver.
3. Yes set interval will fire continuously even when waiting for say the ismaster to return causing potentially multiple hanging queries as it would keep on issuing stalled commands to the servers. Way back  I used setInterval. In practice it caused production issues for people.
4. This design decision was made, too simplify the connection management. The two main things that would have to change to make it work as you outline would be. 1. the worker queue model, 2. authentication would have to be moved down into the pool or some other mechanism would have to designed to ensure any credentials are correctly applied.
5. Possible, I will need to look at it.
6. onAll is for auth and logout.

I'm definitively interested in looking at different designs but I can't promise I will fast-track such a major re-design. I will however be willing to keep a 2.0-experimental branch going for now until we can evaluate if it makes sense to refactor the core completely.

So most of the pain is really concerned with the 1 thread pr socket on MongoDB something that will eventually go away (although I cannot say what the time frame is, could be soon could be in awhile). The plan is to switch to an async or thread multiplexed model some time soon.

However the reason I'm kind of slow on making major surgery is that there are still some outstanding design decisions left on the protocol level that could potentially radically change the driver-server communication in the near future and I would consider doing something in the same motion rather than potentially have to do major surgery twice in a year.

owenallenaz:
For these type of things I'm always in favor of an iterative approach as opposed to a giant overhaul. With this in mind it seems the smallest approach to the socket round robin issue would be the following:

1. I won't touch the onAll, auth, or monitoring components at all. Right now the onAll, capability uses it's own pathway in Server.js which goes through the older pool.get() / pool.getAll() and still calls connection.write(). That pathway still functions normally in my branch and is unchanged by me.

2. I'll update my queue worker to play nice with the reserved monitor socket, ensuring that the inspectServer ismaster check goes straight through to a reserved socket.

With those two things in mind my branch will only be changing the way the server queues up cursors and commands (which are the vast majority of mongo queries). Then in future branches I could clean-up other things, like authing, and onAll, but we'll save for that for another day. How's that sound?

christkv:
more likely to be acceptable but it will take a couple of weeks before I get around to this one as we have other stuff going on that needs my focus right now.

christkv:
Unfortunately it does not pass the mongodb test suite (excluding exhaust). I suspect it needs a fair bit more work to make it work. To test it just repo the mongodb driver and do an npm install of mongodb-core from the 2.0 branch that contains this commit.

owenallenaz:
Stashing progress, I am not complete yet. Still working on getting all of the unit tests to pass. I'm about 50% through the test suite.

christkv:
I'll be interested to see the end result and code review it. There are a bunch of ugly corner cases in the driver due needing to support so many version of mongodb over time.

owenallenaz:
Alright I think I finally have a fully working version of these changes. Overall the general idea is to abstract the management of connections into Pool, so that the users of Pool are oblivious what's going on under the hood, allowing it to be refactored easier. Previously, code all over the place reached into the inner workings of Pool, making refactoring very, very challenging. This is a similar abstraction model as to what is used with Mongos, and ReplSet. In example with ReplSet you simply call .insert() and the abstraction of multiple servers is handled by the inner workers of ReplSet.

1. I've preserved the reserved replSet monitoring connection. It utilizes the last connection in the available connections. capConnections() also will ensure that the last connection is still intact.

2. On my box all unit tests in the mongodb-core repo (verified with travis as well) are passing.

3. On my box all but two of the node-mongodb-native tests are passing. There are two tests which are not passing. On the other hand at least on my machine they aren't passing on the trunk either so I'm not exactly sure what the story is with those. When I comment them out I can get through the whole test suite.

failed tests:

gridfs_stream_tests.js "start/end options for openDownloadStream"
gridfs_tests.js "shouldCorrectlyPerformWorkingFiledWithBigFile" 

4. There is a pathway with CommandResult in server.js:executeSingleOperation where it returns the connection that ran the call. From my reading of the code this is really only used in unit tests, thus I exposed it via an argument in those tests. I could make it appear always, but there is some possiblities of leaking memory if they aren't cleaned up, and that can be tricky to prevent and I wasn't sure it was worth it for a test only feature.

Let me know if I'm still missing something, or if you want to discuss why I made certain decisions etc.

christkv:
2.1.5 is out

chrisckchang:
@owenallenaz you're a gentleman and a scholar.

this is huge! thank you both for this.

PR #74, opened 2016-02-04
auth/sspi: remove annoying console.dir(err)

jonathanong:
coworkers think this is an error. don't need to log it. it would also be nicer if you used `console.error(err.stack)` instead so that we know where the error is coming from, but that's a different issue.


fourpixels:
Oops, I just opened a new issue for that: https://github.com/christkv/mongodb-core/issues/75

Sorry I didn't check here first. Should I close it now or after this pull is merged?

christkv:
fixed in 1.3.1

PR #76, opened 2016-02-10
emit connect event when server reconnects after initial connection failed

vkarpov15:
Re: Automattic/mongoose#1774 . If reconnect is on, server tries to reconnect when initial connection fails, but when it reconnects the server object will be in a broken state because it doesn't have a `wireProtocolHandler`

### DIFF
--- a/lib/topologies/server.js
+++ b/lib/topologies/server.js
@@ -213,6 +213,9 @@ var reconnectServer = function(self, state) {
         count = count - 1;
         // We are done, emit reconnect event
         if(count == 0) {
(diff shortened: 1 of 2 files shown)


christkv:
does the main driver pass replset and single server tests with this change ?

christkv:
merged into 1.3 branch

vkarpov15:
FWIW ran it with the main driver sharding + repl set and tests passed. AFAIK this wouldn't affect MongoClient, just applications that bypass mongoclient and construct servers and dbs directly (like mongoose does).

PR #81, opened 2016-03-10
Add unref

allevo:
https://jira.mongodb.org/browse/NODE-674

### DIFF
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -459,6 +459,21 @@ Connection.prototype.isConnected = function() {
 }
 
 /**
+ * Unref this connection
+ * @method
+ * @return {boolean}
+ */
+Connection.prototype.unref = function() {
+  if (this.connection) this.connection.unref();
+  else {
+    var self = this;
+    this.once('connect', function() {
+      self.connection.unref();
+    });
+  }
+}
+
+/**
  * A server connect event, used to verify that the connection is up and running
  *
  * @event Connection#connect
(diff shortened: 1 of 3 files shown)


christkv:
can you provide a test case or two showing how this will fix your scenario.

allevo:
My show case is described below:
I have a webserver and its the process can handle a unix signal. The handler of this signal (i.e. SIGTERM SIGINT) should stop the process gracefully. As first thing, I refuse all new http connections and then stop all others connections (i.e. memcache and mongodb).

Some problems is raised when the signal handler would stop the process while there's a http request that uses the database has being processing: the `destroy` method on server object destroys the socket used to make the query (a read or write request!). So the http request cannot perform the query and the result is unknown: I don't know if the database has understood the query or if the query is already in the buffer and hasn't been sent to the mongodb server yet.

I would wait that all queries are made to stop the process.

My handler is like this
```
process.on('SIGINT', function() {
  httpServer.close(function() {
    mongoServerConnection.destroy();
  });
});
```

With this pull request, I can write:
```
process.on('SIGINT', function() {
  httpServer.close(function() {
    mongoServerConnection.unref();
  });
});
```
This `unref`ing allows to the process to shut down gracefully: a node process still alive until the stack is empty or a listened is already attached to a event emitter (like a socket).

With this PR, the process has not this (or those) socket(s) opened and dies in peace.

allevo:
Thanks

PR #91, opened 2016-04-11
Fix retry mechanism throwing an error on each retry and not on retry failure.

git-jiby-me:
This PR fixes an issue that, the driver emits an error when each retry fails and doesn't emit an error when retries reach the limit, and driver stops retry.

After this PR, each retry doesn't emit an error, and error is only emitted when retries fails after reaching retry limit.

`reconnectErrorHandler` was being used when an already established
connection errors out, however this should be used instead when a retry
errors out.

### DIFF
--- a/lib/topologies/server.js
+++ b/lib/topologies/server.js
@@ -199,7 +199,7 @@ var reconnectServer = function(self, state) {
     // No more retries
     if(state.currentReconnectRetry <= 0) {
       self.state = DESTROYED;
-      self.emit('error', f('failed to connect to %s:%s after %s retries', state.options.host, state.options.port, state.reconnectTries));
+      self.emit('error', new MongoError(f('failed to connect to %s:%s after %s retries', state.options.host, state.options.port, state.reconnectTries)), self);
     } else {
       setTimeout(function() {
         reconnectServer(self, state);
@@ -225,7 +225,7 @@ var reconnectServer = function(self, state) {
     state.state = CONNECTED;
 
     // Add proper handlers
-    state.pool.on('error', reconnectErrorHandler);
+    state.pool.on('error', errorHandler(self, state));
     state.pool.on('close', closeHandler(self, state));
     state.pool.on('timeout', timeoutHandler(self, state));
     state.pool.on('parseError', fatalErrorHandler(self, state));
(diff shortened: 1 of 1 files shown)


git-jiby-me:
@christkv please take a look.

christkv:
I'm not even sure it all is needed anymore due to the way the pool works. If yo have identified a problem can you provide steps to reproduce ?

git-jiby-me:
@christkv yes sure, i am using `mongodb-native` driver to connect to a single mongodb server. When mongodb server goes offline, i receive an error event on each retry (which doesn't make sense given that we are retrying), and once the retry limit is reached, i don't even get an error event saying that retry failed.

git-jiby-me:
@christkv with this patch, i don't get the error event on each retry failure => which is good, and i get an error event when retry fails out reaching the retry limit.

git-jiby-me:
@christkv Basically without this patch i am unable to know, when server goes offline and the retry fails (after reaching limit).

christkv:
a fix will probably go in a maintenance release in about a month as I'm keeping the driver stable for the next couple of weeks.

these kind of changes can have subtle domino effects on replicaset monitoring and mongos monitoring and will take a while to test and verify.

git-jiby-me:
@christkv Makes complete sense to me. If someone guides me, i could even help with the testing, but i am not aware of the workflow as of now.

Would the maintenance release be published on npm, aka any way i would be able to use it till it lands in a proper release?

christkv:
The maintenance release would be the next version 2.1.17 of the mongodb driver. You cannot easily use it until then without a bunch of package.json messing around.

You can add your additional error handler by going down to the topology.

```js
MongoClient.connect('', function(err, db) {
  db.serverConfig.s.server.on('error', function(err) {})
});
```

git-jiby-me:
@christkv well this patch is not about not able to listen to error event on the db instance, but about the error event on `db.serverConfig.s.server`.

My test case is with 
```
MongoClient.connect('', function(err, db) {
  db.serverConfig.s.server.on('error', function(err) {
    logger.error('MongoDb connection errored', {
      error: err.toString()
    });
  });
});
```
and `serveroptions` set to 
```
{
  "socketOptions": {
      "auto_reconnect": true
  },
  "reconnectTries": 3,
  "reconnectInterval": 1000
}
```

Without the patch (aka mongodb-native 2.1.16) i get this log output
screen shot 2016-04-11 at 11 10 52 pm

With this patch i get this output
screen shot 2016-04-11 at 11 12 12 pm

I get this issue and what the fix changes clearly.

git-jiby-me:
@christkv without the patch, you can see that each retry (which fails as well) is shown as an error event. And for some reason each failed retry error is triggering two error events strangely. And there is no way to know, if the retry failed by reaching the limit.

git-jiby-me:
@christkv  thanks for accepting the PR :)

PR #96, opened 2016-05-17
docs: add explain option to cursor

vkarpov15:
Technically `explain` is a valid option to the cursor constructor, at least against mongodb 3.0, [see example](https://github.com/vkarpov15/mongoose-explain/blob/091832daf622ec93ce97dc2691ce68b5e08a73d4/index.js#L24-L26). Is this formally supported or an accident?

### DIFF
--- a/lib/cursor.js
+++ b/lib/cursor.js
@@ -59,6 +59,7 @@ var Long = require('bson').Long
  * @param {object} [options.transforms=null] Transform methods for the cursor results
  * @param {function} [options.transforms.query] Transform the value returned from the initial query
  * @param {function} [options.transforms.doc] Transform each document returned from Cursor.prototype.next
+ * @param {boolean} [options.explain=false] Explain the query execution rather than return the actual results
  * @param {object} topology The server topology instance.
  * @param {object} topologyOptions The server topology options.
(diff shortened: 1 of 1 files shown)


christkv:
explain is not really a special option on the cursor of mongodb-core. the cursor method on the server/replicaset/mongos class takes a virtual find command that may contain the explain field.

christkv:
passing in explain as an option to the mongodb-core cursor does not do anything

vkarpov15:
Hmm that's funny, I tried tweaking one of the [cursor tests](https://github.com/christkv/mongodb-core/blob/1.3/test/tests/functional/cursor_tests.js) to use `explain` as an option and it definitely does change the result...

christkv:
yes if you tweak the content of the find command it will work but it's not part of the options themselves

vkarpov15:
Ok I'll close this, good to know.

vkarpov15:
FWIW it's a little odd that `explain` is a supported option for `findOne()` and `aggregate()`, but not `find()`. Would you accept a PR to support that?

christkv:
the options for find are deprecated. you are supposed to do

collection.find({a:1}).explain(function(err, r) {})

PR #99, opened 2016-05-25
Pool gets stuck if a connection marked for immediateRelease times out

nbrachet:
since that connection is still in the connectingConnections array, getAll() returns 1 connection and doesn't restart the pool.
However pool.isConnected() returns false since connection.isConnected() returns false and pool.state was left as CONNECTING. All new work gets deferred on the disconnectHandler waiting for reconnection which will never happen.


christkv:
do you have a test case that reproduces this ?

nbrachet:
I don't have one handy, but let me see if I can put one together...

PR #102, opened 2016-05-31
reset socketTimeout which may have been overwritten during a monitoring operation

nbrachet:


christkv:
I'm not sure what you are trying to accomplish the goal is to use the connection timeout setting for the monitoring operations not the socket timeout as we don't want the call to hang.

nbrachet:
Here's the scenario: I have a connectionTimeout set to 5s, and a socketTimeout of 2min.
The current behavior (without the PR) the pool is created with one connection in it, with the correct connectionTimeout then socketTimeout. However as soon as a monitoring operation is scheduled that socket gets the monitoringSocketTimeout -- which defaults to connectionTimeout. 5s later the connection times out, instead of the expected 2min socketTimeout.
That's why I propose to reset the timeout on the socket to socketTimeout when the connection is released back into the pool. That way monitoring operations are executed under the monitoringSocketTimeout, and regular operations under the socketTimeout.

I am confused about why monitoring operations would use the connection timeout, and not the socket timeout? The connection is connected by the time the monitoring operation is scheduled.
That's the second part of this PR which defaults the monitoringSocketTimeout to socketTimeout instead of connectionTimeout.
Note that would incidentally "fix" the problem I describe above, but wouldn't be quite correct -- all you'd have to do is add an explicit monitoringSocketTimout to the mix.

lib/connection/pool.js
```diff
@@ -83,9 +83,9 @@ var Pool = function(options) {
   this.executing = false;
   // Unref pool
   this.unreference = false;
-  // Set the monitoring connection timeout
+  // Set the monitoring socket timeout
   this.monitoringSocketTimeout = typeof options.monitoringSocketTimeout == 'number'
-    ? options.monitoringSocketTimeout : options.connectionTimeout;
+    ? options.monitoringSocketTimeout : options.socketTimeout;
```
christkv: should still be connectionTimeout as you want the monitoring process for a replicaset to timeout as fast as possible if there is a hanging socket write so it can retry.

Issue #105, opened 2016-06-09
Missing connection ID in APM API


laurie71:
The connectionId property of the event object passed to "started" instrumentation event handlers is supposed to always be set, but is often being passed as `undefined`. I traced this back to `Pool.prototype.get()`, having `index` set to NaN. 

I haven't traced the root cause of that, but adding a check to set it to -1 if it is NaN right before it is incremented fixes the problem. I know that's probably not the right fix so I haven't created a pull request. 

Hoping someone can figure out why that property is getting a bad value. I'll keep digging in the mean time and see if I can get to the root cause.

Found in mongodb-core@1.2.31 / mongodb@2.1.21. The same problem occurs using master as of now.

laurie71:
Saw the problem right after posting this issue. It happens when this.connections.length == 0; that causes get() to set this.index to `this.index % 0`. PR incoming.

christkv:
please let me know if this is still a problem in the 2.0.x series of mongodb-core with the mongodb driver 2.2 series

PR #106, opened 2016-06-09
Possible fix for #105

laurie71:
This ensures Pool#index is always valid (or at least never NaN). An undefined connectionId property will still be passed to the APM event handlers if they're called before any connections have been established, but this fixes the problem thats' currently occurring post start-up.


christkv:
closing as the new 2.0 branch have redesigned internals. please reopen if problem persists

PR #107, opened 2016-07-08
Fix read preference tag selection bug

tmichel:
Variable `tags` was not defined so it evaluated to `undefined` and when filtering for tags it always returned an empty array of servers. This caused an infinite retry loop of connecting to mongo.

### DIFF
--- a/lib/topologies/replset.js
+++ b/lib/topologies/replset.js
@@ -911,7 +911,7 @@ var clearCredentials = function(s, ns) {
 var filterByTags = function(readPreference, servers) {
   if(readPreference.tags == null) return servers;
   var filteredServers = [];
-  var tagsArray = !Array.isArray(readPreference.tags) ? [tags] : tags;
+  var tagsArray = !Array.isArray(readPreference.tags) ? [readPreference.tags] : readPreference.tags;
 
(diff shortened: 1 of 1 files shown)


christkv:
not sure if this is still a problem in 2.0 branch, closing for now, please reopen if still an issue

tmichel:
You are correct, this is fixed on the 2.0 branch but dropping support for 1.3 while mongoose still uses a version of `mongodb` package that is dependent on 1.3.x version of `mongodb-core` is makes this feature useless for all mongoose users.

christkv:
I expect mongoose to deliver a update using 2.2.x soon as there are no breaking api changes only internal refactorings.

christkv:
what version of the driver are you using that has this problem

tmichel:
We have mongoose 4.4.20 which trickles down to mongodb-core 1.3.18. But I checked mongoose 4.5.3 still uses an older version of mongodb-core.

christkv:
I'l going to speak with val about doing a release for the 4.4.x and 4.5.x branches to use the new driver as it contains a lot of improvements and also will be the mongodb 3.4 official driver series.

PR #113, opened 2016-07-15
fix: Properly closes connection when it timeout

mlequain:


christkv:
what does this do ? a socket timeout should cause the connection to be closed

mlequain:
The [Node.js documentation](https://nodejs.org/api/net.html#net_event_timeout) specify that a socket timeout doesn't automatically close the connection, it should be closed manually by the user. 

> Event: 'timeout'#
> 
> Added in: v0.1.90
> Emitted if the socket times out from inactivity. This is only to notify that the socket has been idle. The user must manually close the connection.
> 
> See also: socket.setTimeout()

That's why the sockets that timeout are not properly closed and are still open when discarded by your driver, which leads to new connections being created almost continuously since stale ones are never closed properly.

This seems to be happening since this [commit](https://github.com/christkv/mongodb-core/commit/e5d35927b473b5a614c4bd2da9295b18b6e5df3e) but i might be mistaken.

christkv:
fixed and released in 2.2.4

PR #119, opened 2016-07-26
Ensures calls to Pool.prototype.write() maintain their domain

owenallenaz:
Adds domain logic to the primary query entrance point. According to my tests this allowed the domain to remain for insert(), and find() with consistency, including calls routing through getMore() via batchSize() cursors. This is possible due to the pool refactor, previously it regarded domain peppering through the code base.

### DIFF
--- a/lib/connection/pool.js
+++ b/lib/connection/pool.js
@@ -822,7 +822,20 @@ Pool.prototype.write = function(buffer, options, cb) {
     if(cb) cb(new MongoError('pool destroyed'));
     return;
   }
-
+  
+  if (process.domain && typeof cb === "function") {
+    // if we have a domain bind to it
+    var oldCb = cb;
+    cb = process.domain.bind(function() {
+      // v8 - argumentsToArray one-liner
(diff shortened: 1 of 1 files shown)


christkv:
I definitively am not going to allow a nextTick as it will slow throughput in the driver a fair bit. I might make this configurable instead to avoid the additional latency hit performing the callback in a next tick will imply.

christkv:
merged manually, added tests in mongodb-native and made optional via domainsEnabled option on Server/ReplSet/Mongos and MongoClient.connect

owenallenaz:
I'm fine with the setting, but it is important to point out that the nextTick or setImmediate is absolutely essential whenever working with domains. If it is not done, then domains end up leaking if a new domain is entered within the callback of another domain. It  also must be a nextTick and *not* a setImmediate or the like due to the eccentricities of domains. Regarding performance nextTick is barely noticeable because it queues it at the top of the IO loop, meaning that it's async in flow but sync in performance.

Here is the test file I used when creating the pull request, and you can see what happens if we don't bounce off the event loop.

https://gist.github.com/owenallenaz/5111ac4dd1fdc9564e15c25dfc443cd9

PR #134, opened 2016-08-25
fix: don't reconnect multiple times in one connection

dead-horse:
when the network was cut off, the `close` event and `timeout` event will emitted both. https://github.com/christkv/mongodb-core/blob/2.0/lib/connection/pool.js#L318

this will leading to more and more reconnect attempts.

### DIFF
--- a/lib/connection/pool.js
+++ b/lib/connection/pool.js
@@ -277,6 +277,8 @@ function attemptReconnect(self) {
     // If we have failure schedule a retry
     function _connectionFailureHandler(self, event) {
       return function() {
+        if (this._connectionFailHandled) return;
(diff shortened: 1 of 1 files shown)


lib/connection/pool.js
```diff
@@ -277,6 +277,8 @@ function attemptReconnect(self) {
     // If we have failure schedule a retry
     function _connectionFailureHandler(self, event) {
       return function() {
+        if (this._connectionFailHandled) return;
```
dead-horse: not sure if we can use `this.destroyed` to replace `this._connectionFailHandled `

christkv:
merged in manually and applies to other connection failure handler aswell

PR #141, opened 2016-09-30
NODE-798 Driver hangs on count command in replica set with one member

isayme:
`count` should not hangs on when replicat have only one primary but have `secondaryPreferred` preferred.

---
This change is [Reviewable](https://reviewable.io/reviews/christkv/mongodb-core/141)


christkv:
can you please add a test case

isayme:
@christkv test added

PR #147, opened 2016-10-15
Fire callback when topology was destroyed

vkarpov15:
Hi Christian,

Found an annoying issue where a cursor.next() will never fire its callback if the underlying topology ran out of `reconnectTries` or if `autoReconnect` is off. Came up when looking at https://github.com/Automattic/mongoose/issues/4581.


vkarpov15:
Thanks, looking forward to seeing this in the driver :+1:

PR #150, opened 2016-11-10
Dont leak connection.workItems elments when killCursor is called

mdlavin:
When running on mongo 2.4 and 2.6, the killCursor command does not expect a response, but the work is added to the connection.workItems list anyway.  As a result, the connection.workItems array will slowly leak memory as cursors are killed on those versions of mongo.


mdlavin:
@christkv The build failures seem unrelated to my changes. Please let me know if there is something I should do to make the build happier.

walkerx:
？how about mongo 3.0?

mbroadst:
@walkerx are you experiencing a problem with version 3.0 of the server?

walkerx:
@mbroadst YES. workItems are increasing in my case. And I have upgraded to the latest version yesterday.   I will observe if this patch work in these days.

PR #173, opened 2017-03-20
make priviate for the authstore data

michaelnsj:
The purpose is to protect such sensitive information to be accessible from outside. Inspired by https://curiosity-driven.org/private-properties-in-javascript


christkv:
Thanks for the PR. Unfortunately I cannot accept this due having to support previous node versions all the way back to 0.10.x.

PR #177, opened 2017-04-17
set reconnect default to true

wanzhou:


christkv:
reconnections are handled by the mongos class not the individual servers.

wanzhou:
The property `reconnect` is useless.

PR #184, opened 2017-06-02
updates scram.js hi() algorithm to utilize crypto.pbkdf2Sync() for fa…

owenallenaz:
This pull request fixes a few different issues.

1. The previous hi() algorithm was basically a manually written version of crypto.pbkdf2Sync(). My guess is that this library predates the addition of that method in Node v0.9.3. That method in crypto is a C++ native algorithm so it's able to perform the computations dramatically faster. In example on my machine, the previous hi() algorithm would take around 200-300ms while the same pbkdf2Sync version only takes around 12ms.

2. The hi() function is deterministic in the sense that for the same arguments you get the same result back, always. The reason this matter is that when using hi() each worker in the pool will call hi() with the exact same arguments. That means it's doing the same computationally intense call for each worker in the pool even though all of them will utilize the same resulting string. In this fork I'm storing the result of the previous call so that way it only performs the computation once per boot.

3. After the clientFinal generation it appears there was some left-behind code. `// Generate server key` and `// Generate server signature`. Neither of those calls appears to have their result used anywhere in the rest of the function. Removing them appears to have no regression.

If you have questions let me know. Thanks.

### DIFF
--- a/lib/auth/scram.js
+++ b/lib/auth/scram.js
@@ -74,26 +74,19 @@ var xor = function(a, b) {
   return new Buffer(res);
 }
 
-// Create a final digest
+// hiCache stores previous salt creations so it's not regenerated per-pool member
+var _hiCache = {};
+
 var hi = function(data, salt, iterations) {
-  // Create digest
-  var digest = function(msg) {
-    var hmac = crypto.createHmac('sha1', data);
-    hmac.update(msg);
-    return new Buffer(hmac.digest('base64'), 'base64');
-  }
-
-  // Create variables
-  salt = Buffer.concat([salt, new Buffer('\x00\x00\x00\x01')])
-  var ui = digest(salt);
-  var u1 = ui;
-
-  for(var i = 0; i < iterations - 1; i++) {
-    u1 = digest(u1);
-    ui = xor(ui, u1);
-  }
-
-  return ui;
+  var key = `${data}_${data.toString("base64")}_${iterations}`;
+  // check if we've already generated this salt
+  if (_hiCache[key] !== undefined) { return _hiCache[key] }
+  
+  // generate the salt and store it in the cache for the next worker
+  var data = crypto.pbkdf2Sync(data, salt, iterations, 20, "sha1");
+  _hiCache[key] = data;
+  
+  return data;
 }
 
 /**
@@ -232,16 +225,7 @@ ScramSHA1.prototype.auth = function(server, connections, db, username, password,
 
       // Create client final
       var clientFinal = [withoutProof, clientProof].join(',');
-
-      // Generate server key
-      hmac = crypto.createHmac('sha1', saltedPassword);
-      hmac.update(new Buffer('Server Key'))
-      var serverKey = new Buffer(hmac.digest('base64'), 'base64');
-
-      // Generate server signature
-      hmac = crypto.createHmac('sha1', serverKey);
-      hmac.update(new Buffer(authMsg))
-
+      
       //
       // Create continue message
       var cmd = {
--- a/lib/auth/scram.js
+++ b/lib/auth/scram.js
@@ -78,7 +78,7 @@ var xor = function(a, b) {
 var _hiCache = {};
 
 var hi = function(data, salt, iterations) {
-  var key = `${data}_${data.toString("base64")}_${iterations}`;
+  var key = [data, salt.toString("base64"), iterations].join("_");
   // check if we've already generated this salt
   if (_hiCache[key] !== undefined) { return _hiCache[key] }
   
--- a/test/tests/functional/rs_mocks/monitoring_tests.js
+++ b/test/tests/functional/rs_mocks/monitoring_tests.js
@@ -339,7 +339,8 @@ exports['Should correctly prune intervalIds array'] = {
     server.on('connect', function(_server) {
       setTimeout(function() {
         console.dir(_server.intervalIds.length)
-        test.ok(_server.intervalIds.length < 5);
+        // Possible to get 5, 1 for each connection, 1 for the topologyMonitor, and 1 for topologyMonitor executeReconnect before it is has been filtered out from the array by pingServer 
+        test.ok(_server.intervalIds.length <= 5); 
 
         // Destroy mock
         primaryServer.destroy();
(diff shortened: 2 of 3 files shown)


lib/auth/scram.js
```diff
@@ -74,26 +74,19 @@ var xor = function(a, b) {
...
-  for(var i = 0; i < iterations - 1; i++) {
-    u1 = digest(u1);
-    ui = xor(ui, u1);
-  }
+// hiCache stores previous salt creations so it's not regenerated per-pool member
+var _hiCache = {};
 
-  return ui;
+var hi = function(data, salt, iterations) {
+  var key = `${data}_${data.toString("base64")}_${iterations}`;
```
mbroadst: we can't use template strings unfortunately

owenallenaz:
The CI shows failures, but looking at the specific tests that failed I think those are possibly race conditions with the test themselves. Both tests have a timing element to them, and I don't see how my PR could cause a failure in just those scenarios. Feels like the kind of thing where if it just re-run one of those variations it'd pass.

owenallenaz:
Alright I think I got the issues sorted out. The two errors that I corrected were: 

One of them was a timing issue where you were evaling 100ms 5 times. Then we were checking that the time passed was `total > 500` and `total < 1000`. My suspicion is that it's entirely possible to have `total === 500` in that case if it gets to the callback very efficiently. I made both checks `<=` or `>=`.

The other case was checking the number of server intervals. According to my tests with a replset it boots up one interval per server, plus a monitor, plus after success it schedules itself to run again. During the time when it schedules it self to run again, it's entirely possible to have 5 items in the array (1 of which is completed), since the clearing of the array happens in a separate callback chain. By changing the check to be `<= 5` instead of `< 5` it passes. I looked back over the past few months on the CI and see similar failures on the same test so it's not the first time that issues has occurred.

bompus:
@mbroadst Would you mind reviewing this again? If this was implemented, I believe our app startup time would be reduced by about 1 second. The mongodb auth in the driver currently accounts for 20%-30% of our total app init time with 8-10 pool connections having to all go through the slower auth pathway. Thanks!

mbroadst:
@bompus I'm generally okay with this patch, however it would go a long way towards getting it in quickly if you could include a test proving that `crypto.pbkdf2Sync()` and `hi()` were exactly the same. We have many versions of the server to support here, so there is some inherent danger in making this change (even if it seems so small)

owenallenaz:
@mbroadst The tests that currently exist show that auth works correctly. So by replacing the mechanism for hi(), since auth uses that, if the existing tests pass, isn't that proof that my new `hi` mechanism works? If the new variant of `hi` was incorrect, the existing tests should not pass. I can't do a formal unit test to show how a removed function matches a new function, but I did create a simple file you can just execute separately to act as a proof of concept.

```js
var crypto = require("crypto");
var assert = require("assert");

var _hiCache = {};
var newHi = function(data, salt, iterations) {
  var key = [data, salt.toString("base64"), iterations].join("_");
  // check if we've already generated this salt
  if (_hiCache[key] !== undefined) { return _hiCache[key] }
  
  // generate the salt and store it in the cache for the next worker
  var data = crypto.pbkdf2Sync(data, salt, iterations, 20, "sha1");
  _hiCache[key] = data;
  
  return data;
}

var xor = function(a, b) {
  if (!Buffer.isBuffer(a)) a = new Buffer(a)
  if (!Buffer.isBuffer(b)) b = new Buffer(b)
  var res = []
  if (a.length > b.length) {
    for (var i = 0; i < b.length; i++) {
      res.push(a[i] ^ b[i])
    }
  } else {
    for (i = 0; i < a.length; i++) {
      res.push(a[i] ^ b[i])
    }
  }
  return new Buffer(res);
}

var oldHi = function(data, salt, iterations) {
  // Create digest
  var digest = function(msg) {
    var hmac = crypto.createHmac('sha1', data);
    hmac.update(msg);
    return new Buffer(hmac.digest('base64'), 'base64');
  }

  // Create variables
  salt = Buffer.concat([salt, new Buffer('\x00\x00\x00\x01')])
  var ui = digest(salt);
  var u1 = ui;

  for(var i = 0; i < iterations - 1; i++) {
    u1 = digest(u1);
    ui = xor(ui, u1);
  }

  return ui;
}

var tests = [
	["foo", new Buffer("salt1"), 10],
	["foo", new Buffer("something crayz78798321!@*987d98fda"), 100],
	["longer string", new Buffer("1"), 1000],
	["adf987dhfa123hkjdfasdf67a687sdf♦;ï▼|WïΩo", new Buffer("2"), 5]
]

tests.forEach(function(val, i) {
	var oldResult = oldHi.apply(null, val);
	var newResult = newHi.apply(null, val);
	
	assert.strictEqual(oldResult.toString(), newResult.toString());
});

var start = Date.now();
for(var i = 0; i < 100; i++) {
	for(var j = 0; j < tests.length; j++) {
		oldHi.apply(null, tests[j]);
	}
}
console.log("oldHi", Date.now() - start);

var start = Date.now();
for(var i = 0; i < 100; i++) {
	for(var j = 0; j < tests.length; j++) {
		newHi.apply(null, tests[j]);
	}
}
console.log("newHi", Date.now() - start);
```

You can append tests to the bottom by adding your own data strings, with salt buffers and iteration count and this shows that the two results are exactly equal for the `hi` that is in the PR and the `hi` that is in the existing codebase. Also in the file I run each test array for each function 100 times, to show the duration it takes to do the calculation. In my example the return result is:

```
oldHi 1427
newHi 10
```

That helps to show the performance difference between the existing `hi` and the one re-written variant using crypto's native capability.

mbroadst:
@owenallenaz unfortunately the `auth` tests are not run by default in our CI at the moment. I'll have to manually check against all versions of mongodb + node versions, and let you know after that.

owenallenaz:
Sounds good, thank you. I verified the different node versions using my script above and all of them passed. The node docs on the function in question show it existing since Node 0.5.0. I attempted to see if I could get the travis to run mongo in both auth and non-auth mode so that you could but I'm not able to get the test suite to run on my local box. When I run npm test it just does this 

```
> mongodb-core@2.1.12 test /sv/gits/mongodb-core
> node test/runner.js -t functional

/sv/gits/mongodb-core/db/data-27017
(node:12596) DeprecationWarning: 'GLOBAL' is deprecated, use 'global'
(node:12596) DeprecationWarning: 'root' is deprecated, use 'global'
Running tests against MongoDB version `3.4.5`
[purge the directories]
```
If I could get them running, I could see if there's a way to have a test file install auth for that test file, that way you could have the CI test the auth, since doing it manual must be awful.

mbroadst:
thanks for the contribution!

owenallenaz:
Thanks for rolling this on.

PR #186, opened 2017-06-06
[NODE-1021] [NODE-1023] OP_COMPRESSED reception support and mock testing

malexandert:
@sebthedev


lib/connection/connection.js
```diff
@@ -9,7 +9,8 @@ var inherits = require('util').inherits
   , debugOptions = require('./utils').debugOptions
   , Response = require('./commands').Response
   , MongoError = require('../error')
-  , Logger = require('./logger');
+  , Logger = require('./logger')
+  , Snappy = require('snappy');
```
christkv: Snappy must be referenced like bson-ext using require_optional as we cannot rely on the user having or wanting to instal the native lib.

lib/connection/pool.js
```diff
@@ -412,7 +412,7 @@ function messageHandler(self) {
     }
 
     // Log if debug enabled
-    if(self.logger.isDebug()) {
+    if(self.logger.isDebug() || true) {
```
christkv: I think this will always resolve to true meaning the debug message will be printed.

lib/topologies/shared.js
```diff
@@ -56,6 +56,20 @@ function createClientInfo(options) {
   return clientInfo;
 }
 
+function createCompressionInfo(options) {
+  if ("compression" in options && "compressors" in options.compression) {
```
christkv: I don't recommend using in as it can give false positives https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in It might be better to just use indexOf to avoid the side effects of in.

package.json
```diff
@@ -18,7 +18,8 @@
   ],
   "dependencies": {
     "bson": "~1.0.4",
-    "require_optional": "~1.0.0"
+    "require_optional": "~1.0.0",
+    "snappy": "~6.0.0"
```
christkv: Move to devDependencies as it needs to be an optional dependency (all native libs are optional dependencies for the driver). This is critical for windows users that usually don't have a full compiler toolchain installed.

test/mock/lib/request.js
```diff
@@ -26,22 +31,52 @@ Request.prototype.reply = function(documents, options) {
   var startingFrom = typeof options.startingFrom == 'number' ? options.startingFrom : 0;
   var numberReturned = documents.length;
 
+  if ("compression" in options && "compressor" in options.compression) {
```
christkv: See reference about usage of in above

lib/connection/commands.js
```diff
@@ -419,9 +421,46 @@ var Response = function(bson, data, opts) {
   this.responseTo = data[this.index] | data[this.index + 1] << 8 | data[this.index + 2] << 16 | data[this.index + 3] << 24;
   this.index = this.index + 4;
 
-  // Skip op-code field
+  // Read op-code field
+  this.opCode = data[this.index] | data[this.index + 1] << 8 | data[this.index + 2] << 16 | data[this.index + 3] << 24;
   this.index = this.index + 4;
 
+  // If the message we get is OP_COMPRESSED, then we have to uncompress the data 
+  if (this.opCode == 2012) {
```
mbroadst: nit: I would prefer if we didn't use this magic number here (opcode 2012), rather name it as a constant above

lib/connection/commands.js
```diff
@@ -419,9 +421,46 @@ var Response = function(bson, data, opts) {
...
+    this.index = this.index + 4;
+
+    // Read compressor ID
+    var compressorID = data[this.index];
+    this.index = this.index + 1;
+
+    // Read compressed data
+    var compressedData = data.slice(this.index)
+    switch (compressorID) {
+      case 0:
```
mbroadst: nit: same as above with magic numbers here, let's use constants

lib/connection/commands.js
```diff
@@ -419,9 +421,46 @@ var Response = function(bson, data, opts) {
...
+        var uncompressedData = compressedData;
+        break;
+      case 1:
+        var uncompressedData = Snappy.uncompressSync(compressedData)
```
mbroadst: Hmm, unfortunately I don't think this is going to cut it. We will need to use the async versions of `uncompress`/`unzip` instead, it's no good to block the main thread this way. We'll go over possible solutions to this tomorrow.

lib/topologies/server.js
```diff
@@ -247,8 +249,11 @@ var eventHandler = function(self, event) {
       // Query options
       var queryOptions = { numberToSkip: 0, numberToReturn: -1, checkKeys: false, slaveOk: true };
       // Create a query instance
-      var query = new Query(self.s.bson, 'admin.$cmd', {ismaster:true, client: self.clientInfo}, queryOptions);
-      // Get start time
+      if (self.s.compression && self.s.compression.compressors) {
```
mbroadst: it looks like this all could be reduced to:
```
var compressors = (self.s.compression && self.s.compression.compressors) ? self.s.compression.compressors : [];
 var query = new Query(self.s.bson, 'admin.$cmd', {ismaster:true, client: self.clientInfo, compression: compression}, queryOptions);
```

Also, its curious that you seem to be assigning `compressors` to a property called `compression`, and not `self.s.compression`. Is that intentional?

lib/topologies/shared.js
```diff
@@ -56,6 +56,20 @@ function createClientInfo(options) {
   return clientInfo;
 }
 
+function createCompressionInfo(options) {
+  if (options.compression && options.compression.compressors) {
```
mbroadst: nit: I'm an advocate of bailing early. IMHO it helps readability, and reduces an indentation level. e.g.
```
if (!options.compression || !options.compression.compressors) {
  return [];
}

... the rest ...
```

test/mock/lib/request.js
```diff
@@ -1,4 +1,9 @@
-var Long = require('bson').Long;
+var Long = require('bson').Long,
+    Snappy = require('snappy');
+    zlib = require('zlib');
+
+var SNAPPY_ID = 1,
```
mbroadst: ah, looks like we have constants here. When you use them in the module itself, you can drop these and share the code

test/tests/functional/single_mocks/compression_tests.js
```diff
@@ -0,0 +1,388 @@
+exports['should error is an invalid compressor is specified'] = {
```
mbroadst: "should error _if_..."

lib/topologies/server.js
sebthedev: Thanks for that more simple version – I've implemented it now.

We accept a `compression` option (with properties `compressors` and `zlibCompressionLevel`) from the user, but we then need to flatten it down to just an array of compressors for the handshake with the server (according to the spec). Not sure if this answers your question…?

lib/connection/commands.js
```diff
@@ -223,6 +229,9 @@ Query.getRequestId = function() {
   return ++_requestId;
 }
 
+Query.prototype.fromBin = function(bson) {
```
mbroadst: are you using this?

lib/connection/commands.js
```diff
@@ -485,6 +471,8 @@ Response.prototype.parse = function(options) {
     promoteBuffers: promoteBuffers
   };
 
+  this.index = 20;
```
mbroadst: what is this magic number?

lib/connection/connection.js
```diff
@@ -7,16 +7,20 @@ var inherits = require('util').inherits
...
+  , zlib = require('zlib')
+  , Snappy = require('snappy');
 
 var _id = 0;
 var debugFields = ['host', 'port', 'size', 'keepAlive', 'keepAliveInitialDelay', 'noDelay'
   , 'connectionTimeout', 'socketTimeout', 'singleBufferSerializtion', 'ssl', 'ca', 'crl', 'cert'
   , 'rejectUnauthorized', 'promoteLongs', 'promoteValues', 'promoteBuffers', 'checkServerIdentity'];
 var connectionAccounting = false;
 var connections = {};
+var OP_COMPRESSED = 2012;
```
mbroadst: is there a way to share this definition? It's being defined twice

lib/connection/connection.js
```diff
@@ -205,6 +209,47 @@ var closeHandler = function(self) {
   }
 }
 
+var decompress = function(compressorId, compressedData, callback) {
+  var decompressors = [
```
mbroadst: nit: this all looks a little too clever imho. At very least lets build the supported decompressor list outside this method, so we're not building the array every time `decompress` is called?

lib/connection/connection.js
```diff
@@ -205,6 +209,47 @@ var closeHandler = function(self) {
...
+    function(compressedData, callback) {
+      callback(null, compressedData)
+    },
+    Snappy.uncompress,
+    zlib.unzip
+  ];
+
+  if (compressorId < 0 || compressorId > decompressors.length - 1) {
+    throw new Error('Server sent message compressed using an unsupported compressor');
```
mbroadst: Include the compressor id in the message to help the user identify what's gone wrong

lib/connection/connection.js
```diff
@@ -205,6 +209,47 @@ var closeHandler = function(self) {
...
+  }
+
+  // Pass the compressed date to the appropriate decompressor
+  decompressors[compressorId](compressedData, function(err, decompressedData) {
+    if (err) {
+      console.log(err)
```
mbroadst: two issues:
  - you're using `console.log` here, instead of the logger
  - when the error occurs you should be calling back with `callback(err, null)` and returning. Presently you just print out the error, and then return null/undefined value for a success path.

lib/connection/connection.js
```diff
@@ -205,6 +209,47 @@ var closeHandler = function(self) {
...
+
+var emitMessageHandler = function (message, self) {
+  var msgHeader = parseHeader(message);
+  if (msgHeader.opCode == OP_COMPRESSED) {
+    msgHeader.fromCompressed = true;
+    msgHeader.opCode= message.readInt32LE(16);
+    msgHeader.length = message.readInt32LE(20);
+    var compressorId = message[24];
+
+    decompress(compressorId, message.slice(25), function(decompressedData) {
```
mbroadst: magic number `25` === ??

lib/connection/connection.js
```diff
@@ -228,6 +273,7 @@ var dataHandler = function(self) {
           // Slice the overflow into a new buffer that we will then re-parse
           data = data.slice(remainingBytesToRead);
 
+          console.log("dataHandler")
```
mbroadst: accidentally left in this debug message

lib/connection/connection.js
```diff
@@ -236,8 +282,9 @@ var dataHandler = function(self) {
             self.sizeOfMessage = 0;
             self.bytesRead = 0;
             self.stubBuffer = null;
-            // Emit the buffer
-            self.messageHandler(new Response(self.bson, emitBuffer, self.responseOptions), self);
+
+            emitMessageHandler(emitBuffer, self);
```
mbroadst: nit: its more idiomatic to include `self` as the first argument to a method

lib/connection/utils.js
```diff
@@ -78,9 +78,20 @@ var retrieveBSON = function() {
   return BSON;
 }
 
+// Parses the header of a wire protocol message
+var parseHeader = function(message) {
```
mbroadst: question: is it worth attempting to check the bounds of the passed in message here? Or is that happening at a higher level? afaict, you just assume there are 16 bytes here which could fail in an unexpected fashion.  If you do end up bounds checking, then please include a unit test for this.

lib/topologies/shared.js
```diff
@@ -56,6 +56,21 @@ function createClientInfo(options) {
...
 }
 
+function createCompressionInfo(options) {
+  if (!options.compression || !options.compression.compressors) {
+    return [];
+  }
+
+  // Check that all supplied compressors are valid
+  options.compression.compressors.forEach(function(compressor) {
+    if (compressor != 'snappy' && compressor != 'zlib') {
```
mbroadst: `!==` vs `!=`

lib/wireprotocol/shared.js
```diff
@@ -22,6 +25,18 @@ var getReadPreference = function(cmd, options) {
   return readPreference;
 }
 
+
+var writeReply = function(bson) {
```
mbroadst: what's going on with this method? it doesn't return anything

test/mock/lib/request.js
```diff
@@ -135,6 +183,74 @@ Response.prototype.toBin = function() {
...
+  // Write response flags
+  writeInt32(dataToBeCompressedHeader, 0, this.uncompressedResponse.responseFlags)
+  writeInt64(dataToBeCompressedHeader, 4, this.uncompressedResponse.cursorId)
+  writeInt32(dataToBeCompressedHeader, 12, this.uncompressedResponse.startingFrom)
+  writeInt32(dataToBeCompressedHeader, 16, this.uncompressedResponse.numberReturned)
+
+  var dataToBeCompressed = Buffer.concat([dataToBeCompressedHeader, dataToBeCompressedBody])
+
+  // Compress the data
+  var compressorFunctions = [
```
mbroadst: less important because its in the testing, but again we are constructing this array _every time_ you compress something

test/tests/functional/server_tests.js
```diff
@@ -127,6 +127,42 @@ exports['Should correctly connect server to single instance and execute insert']
   }
 }
 
+exports['Should correctly connect server to single instance and execute insert (now with compression)'] = {
+  metadata: { requires: { topology: ["single", "snappyCompression"] } },
```
mbroadst: looks like this environment was accidentally depended upon here, but not defined in the runner

lib/connection/connection.js
mbroadst: needs a test

lib/connection/commands.js
```diff
@@ -485,6 +459,9 @@ Response.prototype.parse = function(options) {
     promoteBuffers: promoteBuffers
   };
 
+  // Position within the message body at which documents start
+  this.index = 20;
```
mbroadst: nooow I know why this was bothering me 😄 

So, you pre-parse the header and pass in the msgHeader and msgBody to this class. So you don't need to initialize the `this.index = 20`, because you really should be setting `this.data = msgBody` and keeping the index at 0. `this.raw` has to equal `msgHeader + msgBody`.

.travis.yml
```diff
@@ -1,13 +1,11 @@
...
+sudo: required
+dist: trusty
 language: node_js
+
 node_js:
-  - "0.10"
-  - "0.12"
   - "4"
   - "6"
-  - "7"
```
christkv: Don't forget to add 8

package.json
```diff
@@ -18,7 +18,7 @@
   ],
   "dependencies": {
     "bson": "~1.0.4",
-    "require_optional": "~1.0.0"
+    "require_optional": "github:christkv/require_optional#master"
```
christkv: Published version 1.0.1 to npm

lib/connection/connection.js
```diff
@@ -7,9 +7,15 @@ var inherits = require('util').inherits
...
+  , parseHeader = require('./utils').parseHeader
+  , compressorIDs = require('./utils').compressorIDs
   , Response = require('./commands').Response
   , MongoError = require('../error')
-  , Logger = require('./logger');
+  , Logger = require('./logger')
+  , zlib = require('zlib')
+  , Snappy = require('./utils').retrieveSnappy()
+  , OP_COMPRESSED = require('./commands').OP_COMPRESSED
+  , MESSAGE_HEADER_SIZE = require('../wireprotocol/shared').MESSAGE_HEADER_SIZE;
```
mbroadst: it kind of feels like all of the op constants should live in `wireprotocol/shared`. e.g. `OP_COMPRESSED`, `OP_QUERY`, etc

sebthedev: Sure, I can move them.

PR #187, opened 2017-06-07
Use actual server type in standalone SDAM events:

durran:
Currently the server type inside the topologyDescriptionChanged event
for a single topology is 'Standalone', which is unexpected. According to
the SDAM specification a TopologyDescription has a list of
ServerDescriptions, which had a type field indicating the type of server
it is and does not state that the server type never changes if the
topology is single.

My expectation is that a single connection to a replica set primary, for
example, would give me a topology type of 'Single' and a server type of
'RSPrimary', not 'Standalone' like it currently does.

The serverDescriptionChanged event properly has the type set.

I could not get the tests running locally - they just stop after the first step `[purge the directories]` - hence I have not added tests but the existing sdam server test should still pass.

### DIFF
--- a/lib/topologies/server.js
+++ b/lib/topologies/server.js
@@ -287,7 +287,7 @@ var eventHandler = function(self, event) {
         if(!self.s.inTopology) {
           // Emit topology description changed if something listening
           sdam.emitTopologyDescriptionChanged(self, {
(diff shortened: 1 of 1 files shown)


christkv:
I guess we are missing a test for this particular case, mind adding one.

mbroadst:
I'll merge this in myself, sorry github is being silly

PR #193, opened 2017-06-14
[NODE-1016] [NODE-1027] Completing Wire Protocol Compression support

sebthedev:
These changes complete the support in mongodb-core for full bi-directional wire protocol compression.


lib/connection/utils.js
```diff
@@ -78,9 +78,44 @@ var retrieveBSON = function() {
...
+      uncompress: noSnappyWarning,
+      compressSync: noSnappyWarning,
+      uncompressSync: noSnappyWarning
+    }
+  }
+  return snappy;
+}
+
+// Parses the header of a wire protocol message
+var parseHeader = function(message) {
```
mbroadst: does this belong in `mongodb-core/lib/wireprotocol/shared.js` ?

lib/connection/pool.js
```diff
@@ -907,6 +925,25 @@ Pool.prototype.destroy = function(force) {
   checkStatus();
 }
 
+// Facilitate compressing a message using an agreed compressor
+var compress = function(self, dataToBeCompressed, callback) {
```
mbroadst: it's starting to feel like we might need a separate file in `mongodb-core/lib/wireprotocol` called `compression.js` which encapsulates all this code that relates to compression. Thoughts? (`compress`, `decompress`, `compressorIds`, `uncompressibleCommands`, etc)

lib/connection/pool.js
```diff
@@ -957,64 +994,118 @@ Pool.prototype.write = function(commands, options, cb) {
...
-  var buffer = null
+  // Ensure commands is an array
+  if (!Array.isArray(commands)) {
+    commands = [commands];
+  }
 
-  if(Array.isArray(commands)) {
-    buffer = [];
+  // For each command, create a promise while we create the command to send to the server
+  var commandsPromises = []
```
mbroadst: This sort of dramatically complicates the code imho. The original code is short and sweet:
```
  if(Array.isArray(commands)) {
    buffer = [];

    for(var i = 0; i < commands.length; i++) {
      buffer.push(commands[i].toBin());
    }

    // Get the requestId
    operation.requestId = commands[commands.length - 1].requestId;
  } else {
    operation.requestId = commands.requestId;
    buffer = commands.toBin();
  }
```

Whereas the refactoring here introduces a lot of compression-specific information, as well as forcing everything to use promises - which is generally a no-no in "hot paths", considering the overhead needed to use them. Promises are generally relegated to user-facing API, as their greatest strength is that of readability. 

It looks like the complicated part of this code is building the payload for an `operation`. What I would do is keep as much the same as possible between the sync/async versions of building the operation (things like this: https://github.com/christkv/mongodb-core/blob/2.0/lib/connection/pool.js#L991-L999), isolate the pieces that need to potentially be async and "branch" there - similar to what was done with `decompress`. The pseudocode might look like:

```
  operation.promoteLongs = typeof options.promoteLongs == 'boolean' ? options.promoteLongs : true;
  operation.promoteValues = typeof options.promoteValues == 'boolean' ? options.promoteValues : true;
  operation.promoteBuffers = typeof options.promoteBuffers == 'boolean' ? options.promoteBuffers : false;
  operation.raw = typeof options.raw == 'boolean' ? options.raw : false;
  operation.immediateRelease = typeof options.immediateRelease == 'boolean' ? options.immediateRelease : false;
  operation.documentsReturnedIn = options.documentsReturnedIn;
  operation.command = typeof options.command == 'boolean' ? options.command : false;
  operation.fullResult = typeof options.fullResult == 'boolean' ? options.fullResult : false;
  operation.noResponse = typeof options.noResponse == 'boolean' ? options.noResponse : false;
  ... lots of other operation stuff ....

  if (needToCompress) {
     compress(operation, (err, op) => {
        this.queue.shift(op);
        tryToExecute();
     });
  } else {
    this.queue.shift(op);
    tryToExecute();
  }
```

sebthedev: This improvement makes sense to me!

However, I think that the method you are proposing will only work if `commands` has length one (see https://github.com/christkv/mongodb-core/blob/2.0/lib/connection/pool.js#L973-L985). If there is more than one command, then each of these commands needs to be individually transformed into an OP_COMPRESSED message, which would require multiple async compression tasks. The only tidy way I can think to handle this is through `Promise.all()`.

That linked section of code from 2.0 appears to suggest that sometimes it is possible for more than one command to be sent at the same time. However, I just ran the tests with a console log and I can't see any moment during the tests when more than one command is sent at the same time.

Do you know whether multiple commands are ever actually sent? If multiple commands _are_ ever sent, how would you advise I best wait for the multiple async compression tasks to complete?

lib/connection/pool.js
```diff
@@ -918,6 +925,56 @@ Pool.prototype.destroy = function(force) {
   checkStatus();
 }
 
+// Prepare the buffer that Pool.prototype.write() uses to send to the server
+var serializeCommands = function(self, commands, result, callback) {
+  // Base case when there are no more commands to serialize
+  if (commands.length === 0) return callback(null, result);
+
+  // Pop off the zeroth command and serialize it to
```
mbroadst: serialize it to... ?

sebthedev: Sorry, I should have caught that. Now fixed.

lib/connection/pool.js
```diff
@@ -1007,25 +1049,49 @@ Pool.prototype.write = function(commands, options, cb) {
...
+        _execute(self)();
+      });
+    }
+  })
+
+}
+
+// Return whether a command contains an uncompressible command term
+// Will return true if command contains no uncompressible command terms
+var checkForUncompressibleCommands = function(command) {
```
mbroadst: I feel like this entire method could be rewritten in a line:
```
return uncompressibleCommands.some(function(cmd) { return command.query.hasOwnProperty(cmd); });
```

sebthedev: @malexandert I think you wrote this function originally – what do you want to do about this suggestion?

malexandert: I think both ways are equally readable, but Matt's way is definitely shorter, so we should change it

PR #195, opened 2017-06-29
Fix issue in Response.parse when the underlying cursor has been closed

herbrandson:
This PR resolves a bug when calling `cancel` on a `cursor` while using `forEach` to iterate over a result set. In that situation, the result of `deserialize` on commands.js line #502 is something like
```
{ ok: 0,
  errmsg: 'Cursor not found, cursor id: 38233323205',
  code: 43 }
```

I'm not sure if this is the best solution, but it's a possible solution :)

### DIFF
--- a/lib/connection/commands.js
+++ b/lib/connection/commands.js
@@ -501,6 +501,11 @@ Response.prototype.parse = function(options) {
     // Deserialize but keep the array of documents in non-parsed form
     var doc = this.bson.deserialize(document, _options);
 
+    // Make sure we have a cursor
+    if (!doc.cursor) {
+      return;
+    }
+
     // Get the documents
     this.documents = doc.cursor[documentsReturnedIn];
     this.numberReturned = this.documents.length;


mbroadst:
@herbrandson could you possibly provide a code snippet exemplifying the situation you're trying to solve here?

herbrandson:
Sure thing. Our use case is that we have a websocket based api. Clients connect to this api and request a very large dataset (i.e. millions or results) which we then begin streaming to them. While the we are streaming the results, the client can ask for a new/different dataset. In this case we cancel the previous request (i.e call `cursor.close()`) and start a new cursor to stream the new results. A simplified version of the code is below.

```
function streamCursor(cursor, cancelToken) {
	return new Promise((resolve, reject) => {

		cancelToken.onCancel(() => cursor.close());

		cursor.forEach(processItem, end);

		function processItem(item) {
			// push the data to a websocket here
		}

		function end(err) {
			if (err) {
				reject(err);
			} else {
				resolve();
			}
		}
	});
}
```

The problem is that whenever this happens there is an unhandled exception in mongodb-core. This patch is attempting to resolve this issue.

herbrandson:
In case it's helpful, the error messages we see are either
`TypeError: Cannot read property 'firstBatch' of undefined`
or
`TypeError: Cannot read property 'nextBatch' of undefined`

mbroadst:
@herbrandson can you provide more detail about your actual query, and perhaps what you're doing in `processItem`, I'm unable to simulate this error on my end. For instance, it looks like you're probably using `raw: true` in your query? Here's my existing test, perhaps we could work on improving it together:

```
exports['Should gracefully handle a closed cursor with incoming responses'] = {
  metadata: {
    requires: { topology: ["single", "replicaset", "sharded"] }
  },

  test: function(configuration, test) {
    var Server = require('../../../lib/topologies/server'),
        bson = require('bson');

    var server = new Server({
      host: configuration.host, port: configuration.port, bson: new bson()
    });

    var ns = f("%s.killedCursor", configuration.db);
    var docs = [];
    for (var i = 0; i < 1000; ++i) docs.push({ a: i  });

    server.on('connect', function(_server) {
      _server.insert(ns, docs, {
        writeConcern: { w: 1 }, ordered: true
      }, function(err, results) {
        test.equal(null, err);
        test.equal(docs.length, results.result.n);

        var cursor = _server.cursor(ns, {
          find: ns,
          query: {},
          batchSize: 5,
          raw: true
        });

        setTimeout(function() { cursor.kill(); }, 100);

        // simulate a forEach
        function each(callback) {
          cursor.next(function(err, d) {
            if (err) return callback(err);
            if (d === null) return callback();
            each(callback);
          });
        }

        each(function(err) {
          _server.destroy();
          test.done();
        });
      });
    });

    server.connect();
  }
}
```

mbroadst:
@herbrandson orthogonal to trying to solve the bug here, it seems like you might want to use [streams](https://mongodb.github.io/node-mongodb-native/api-generated/cursor.html#stream) for your particular case here.

herbrandson:
First, thanks a ton for looking into this! I really appreciate it. 

You are correct, we are using `raw: true`. We send the data over the websocket using a binary format (flatbuffers). We ended up writing a SAX style parse for processing the raw BSON because there was significant overhead (both CPU and GC related) in first converting records to JSON. That's mostly what is happening in `processItem`. I may be able to send you some of that code, but I doubt it would be very helpful. It's mostly looping over the raw `buffer` response.

Other things that may or may not be helpful to know:
1. Documents can be somewhat large. Maybe up to around ~50k
2. We use a batch size of 200
3. The queries include a projection to only return certain fields
4. The queries include a `$geoIntersects` filter

Here's an updated test that should be a little closer to what I'm doing.

```
exports['Should gracefully handle a closed cursor with incoming responses'] = {
  metadata: {
    requires: { topology: ["single", "replicaset", "sharded"] }
  },

  test: function(configuration, test) {
    var Server = require('../../../lib/topologies/server'),
        bson = require('bson');

    var server = new Server({
      host: configuration.host, port: configuration.port, bson: new bson()
    });

    var ns = f("%s.killedCursor", configuration.db);
    var docs = [];
    for (var i = 0; i < 1000; ++i) docs.push({ a: i  });

    server.on('connect', function(_server) {
      _server.insert(ns, docs, {
        writeConcern: { w: 1 }, ordered: true
      }, function(err, results) {
        test.equal(null, err);
        test.equal(docs.length, results.result.n);

        var cursor = _server.cursor(ns, {
          find: ns,
          query: {},
          batchSize: 5,
          raw: true
        });

        setTimeout(function() { cursor.kill(); }, 100);

        cursor.forEach(processItem, end);

        function processItem(d) {
          // maybe assert "d" or something here?
        }

        function end(err) {
          _server.destroy();
          test.done();
        };
      });
    });

    server.connect();
  }
}
```

Thanks for the heads up on the `streams` api. Would this still work with `raw: true`?

herbrandson:
Just noticed one other thing. I'm using `cursor.close()` instead of `cursor.kill()`. Maybe that's related?

Here's another approach to the test that may be worth trying...
```
exports['Should gracefully handle a closed cursor with incoming responses'] = {
  metadata: {
    requires: { topology: ["single", "replicaset", "sharded"] }
  },

  test: function(configuration, test) {
    var Server = require('../../../lib/topologies/server'),
        bson = require('bson');

    var server = new Server({
      host: configuration.host, port: configuration.port, bson: new bson()
    });

    var ns = f("%s.killedCursor", configuration.db);
    var docs = [];
    for (var i = 0; i < 1000; ++i) docs.push({ a: i  });

    server.on('connect', function(_server) {
      _server.insert(ns, docs, {
        writeConcern: { w: 1 }, ordered: true
      }, function(err, results) {
        test.equal(null, err);
        test.equal(docs.length, results.result.n);

        var batchSize = 200;
        var batchCount = 0;

        var cursor = _server.cursor(ns, {
          find: ns,
          query: {},
          batchSize: batchSize,
          raw: true
        });

        cursor.forEach(processItem, end);

        function processItem(d) {
          if (++batchCount >= batchSize) {
            cursor.close();
          }
        }

        function end(err) {
          _server.destroy();
          test.done();
        };
      });
    });

    server.connect();
  }
}
```

herbrandson:
Hmmm, one more thing I just noticed. We're getting the cursor in a slightly different way. We actually get the collection first using something like `_server.collection('collection_name', {raw: true})` and then get the cursor by calling `.find` on the collection.

Also, we add the following flag to the cursor
```
cursor.addCursorFlag('noCursorTimeout', true);
```

I don't know if any of that makes a difference, but I thought I'd pass it along just in case :)

herbrandson:
@mbroadst where you able to reproduce? Anything else I can do to help?

mbroadst:
@herbrandson sorry I haven't had time to try again, we're a little busy implementing new server features over here 😄 .  It would help if you could give me a snippet of the actual code you're using to trigger this (from the top level `node-mongodb-native` perspective). The reason our code looks different is because I was trying to simulate this only with `mongodb-core`.  Thanks for your patience.

daprahamian:
hi @herbrandson. We believe that we fixed this issue in the 3.x version of the driver. Are you able to reproduce this issue on 3.x?

herbrandson:
@daprahamian It's been a while since this happened. I know it was resolved. IIRC, the root cause was using a DNS value to connect to the mongos instance that was using a non-sticky round-robin scheme.

PR #196, opened 2017-07-01
refactor(assign): move assign to shared utils, replace extend

mbroadst:
We already include a polyfill for `Object.assign` in the shared
topology utils, so this change moves that to a more centralized
location and replaces all inline implementations of a similar
`extend` method.


christkv:
LGTM

PR #198, opened 2017-07-18
NODE-1080: Create MongoNetworkError

sebthedev:
This PR creates the new MongoNetworkError class of errors. Having MongoNetworkError allows us to identify which errors are resumable errors for the purposes of Change Streams (see [NODE-1057](https://jira.mongodb.org/browse/NODE-1057)).


lib/network_error.js
```diff
@@ -0,0 +1,44 @@
+"use strict";
```
mbroadst: Is there a reason to have this live in its own file? Afaict, it should just live inside `error` and probably actually subclass `MongoError` itself (this is a specialization of that type)

sebthedev: I was unsure how to best do this. I'll research subclassing.

lib/error.js
```diff
@@ -1,44 +1,65 @@
...
+  } else {
+    if (typeof message === 'string') {
+      this.message = message;
+    } else {
+      this.message = message.message || message.errmsg || message.$err || 'n/a';
+      for (var name in message) {
+        this[name] = message[name];
+      }
+    }
+    Error.captureStackTrace(this, MongoError);
```
mbroadst: also I think we _always_ want to capture the stack trace. If an error is passed in, you should include that as like an originating error, it shouldn't _become_ this error

mbroadst: actually forget that, I guess that is the original intent...

PR #207, opened 2017-08-06
refactor(wire-protocol): remove support for pre-2.6 servers

mbroadst:
We no longer support pre-2.6 servers (they are EOL), so it makes sense to start pruning some of the dead code before adding new 3.6 features.


mbroadst:
@christkv added that check (when you're back around)

PR #214, opened 2017-08-16
[NODE-1096] Converting tests to run with mongodb-test-runner, round 2

malexandert:
@mbroadst here are the rest of the tests

### DIFF
--- a/test/tests/functional/basic_replset_server_auth_tests.js
+++ b/test/tests/functional/basic_replset_server_auth_tests.js
@@ -1,59 +1,66 @@
-"use strict";
+'use strict';
 
-var f = require('util').format,
-  locateAuthMethod = require('./shared').locateAuthMethod,
-  executeCommand = require('./shared').executeCommand;
+var expect = require('chai').expect,
+    f = require('util').format,
+    locateAuthMethod = require('./shared').locateAuthMethod,
+    executeCommand = require('./shared').executeCommand,
+    ReplSet = require('../../../lib/topologies/replset'),
+    Connection = require('../../../lib/connection/connection');
 
 var setUp = function(configuration, options, callback) {
   var ReplSetManager = require('mongodb-topology-manager').ReplSet;
 
   // Check if we have any options
-  if(typeof options == 'function') callback = options, options = null;
+  if (typeof options === 'function') {
+    callback = options;
+    options = null;
+  }
 
   // Override options
-  if(options) {
-    var rsOptions = options;
+  var rsOptions;
+  if (options) {
+    rsOptions = options;
   } else {
-    var rsOptions = {
+    rsOptions = {
       server: {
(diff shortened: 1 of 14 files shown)


test/tests/functional/operation_example_tests.js
```diff
@@ -1,873 +1,850 @@
...
-        writeConcern: {w:1}, ordered:true
-      }, function(err, results) {
-        test.equal(null, err);
-        test.equal(1, results.result.n);
-
-        _server.destroy();
-        test.done();
-      });
-    });
+describe.skip('Mongos operation example tests', function() {
```
malexandert: not sure what to do here, since there is no `mongos` topology to setup, and running the tests under the `sharded` topology causes them to error

test/tests/functional/basic_replset_server_auth_tests.js
```diff
@@ -71,605 +78,283 @@ var setUp = function(configuration, options, callback) {
...
-              host: 'localhost'
-            , port: 31001
-          }], {
-            setName: 'rs'
-          });
+};
 
-          server.on('connect', function(_server) {
-          });
+describe.skip('Basic replica set server auth tests', function() {
```
malexandert: also skipping these tests because of that `keyfile` thing I sent you

test/tests/functional/operation_example_tests.js
mbroadst: that's fine for now, I'm not sure they were being run for all releases anyway

mbroadst:
@malexandert Hmm looks like the build is failing for a non-obvious reason: https://travis-ci.org/christkv/mongodb-core/jobs/265363148#L555

test/tests/functional/server_tests.js
```diff
@@ -848,19 +846,19 @@ describe('Server tests', function() {
       Connection.enableConnectionAccounting();
 
       this.configuration.manager.restart(true).then(function() {
-        locateAuthMethod(this.configuration, function(err, method) {
+        locateAuthMethod(self.configuration, function(err, method) {
           expect(err).to.be.null;
 
           // Attempt to connect
-          executeCommand(this.configuration, 'admin', {
+          executeCommand(self.configuration, 'admin', {
```
mbroadst: wow how was this running at all before??

mbroadst:
Generally LGTM after you get the error above fixed. Ready to merge after a successful CI run.

test/tests/functional/server_tests.js
malexandert: Didn't catch this earlier because they're `auth` tests, so they didn't run at all when I tested `single`, `replicaset`, and `sharded` environments

malexandert:
@mbroadst the CI run is passing now. There was a bug in the test runner causing the non-obvious failure, and I added `.skip` to a couple of flaky pool tests

PR #218, opened 2017-09-24
feat(sessions): track `logicalSessionTimeoutMinutes` for sessions

mbroadst:
This adds required tracking of `logicalSessionTimeoutMinutes` from ismaster responses according to the new sections in the SDAM spec:
  - https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#logical-session-timeout
  - https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#logicalsessiontimeoutminutes

### DIFF
--- a/lib/topologies/replset_state.js
+++ b/lib/topologies/replset_state.js
@@ -72,6 +72,8 @@ var ReplSetState = function(options) {
     topologyType: 'Unknown',
     servers: []
   };
+
+  this.logicalSessionTimeoutMinutes = undefined;
 };
 
 inherits(ReplSetState, EventEmitter);
@@ -261,6 +263,21 @@ ReplSetState.prototype.update = function(server) {
 
     // Set the topology
     return false;
+  }
+
+  // Update logicalSessionTimeoutMinutes
+  if (ismaster.logicalSessionTimeoutMinutes !== undefined) {
+    if (
+      self.logicalSessionTimeoutMinutes === undefined ||
+      ismaster.logicalSessionTimeoutMinutes === null
+    ) {
+      self.logicalSessionTimeoutMinutes = ismaster.logicalSessionTimeoutMinutes;
+    } else {
+      self.logicalSessionTimeoutMinutes = Math.min(
+        self.logicalSessionTimeoutMinutes,
+        ismaster.logicalSessionTimeoutMinutes
+      );
+    }
   }
 
   //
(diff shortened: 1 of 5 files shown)


lib/topologies/replset_state.js
```diff
@@ -263,6 +265,21 @@ ReplSetState.prototype.update = function(server) {
...
 
+  // Update logicalSessionTimeoutMinutes
+  if (ismaster.logicalSessionTimeoutMinutes !== undefined) {
+    if (
+      self.logicalSessionTimeoutMinutes === undefined ||
+      ismaster.logicalSessionTimeoutMinutes === null
+    ) {
+      self.logicalSessionTimeoutMinutes = ismaster.logicalSessionTimeoutMinutes;
+    } else {
+      self.logicalSessionTimeoutMinutes = Math.min(
```
kmahar: so it says in specs "it MUST set TopologyDescription.logicalSessionTimeoutMinutes to the smallest logicalSessionTimeoutMinutes value among all ServerDescriptions of known ServerType". 

if this is just taking the minimum of the master and itself, will that accomplish that? (maybe it is communicated to other nodes through the master?)

mbroadst: Yeah there was definitely two ways to go about this: 
  - track each incoming `logicalSessionTimeoutMinutes` from ismaster responses for connected replset members, and then on request of the value iterate through all and find the lowest one

  - track each `logicalSessionTimeoutMinutes` as they come in, and only ever store it once

Unless you're seeing something wrong with the logic, the intent here is to "pre-compute" the smallest known value

test/tests/unit/sessions_tests.js
```diff
@@ -0,0 +1,253 @@
...
+            done();
+          });
+
+          replset.connect();
+        });
+      }
+    }
+  );
+
+  it('should set `logicalSessionTimeoutMinutes` to `null` if any incoming server is `null`', {
```
kmahar: this looks identical to previous test, minus the secondary's logicalSessionTimeoutMinutes value and the replSet's expected value. maybe you could combine them and just loop over values to use for the secondary and the expected value?

mbroadst: Yeah, totally. I'm actually refactoring this presently for the cluster time gossiping stuff, I think we should do the refactor there instead.

lib/topologies/replset_state.js
kmahar: ahhh ok I see! I was thinking that only the master itself sends ismaster responses. but now that it look into it I see that all the members send them. makes more sense

PR #219, opened 2017-09-27
Cluster time gossiping

mbroadst:


lib/connection/pool.js
```diff
@@ -542,6 +544,12 @@ function messageHandler(self) {
           return handleOperationCallback(self, workItem.cb, new MongoError(err));
         }
 
+        // Look for clusterTime, and update it if necessary
+        if (message.documents[0] && message.documents[0].hasOwnProperty('$clusterTime')) {
+          const $clusterTime = message.documents[0].$clusterTime;
```
kmahar: what is the rationale behind doing this as opposed to just directly assigning `self.topology.clusterTime = message.documents[0].$clusterTime`? I don't see $clusterTime used again

lib/connection/pool.js
```diff
@@ -1134,6 +1142,16 @@ Pool.prototype.write = function(commands, options, cb) {
   // Get the requestId
   operation.requestId = commands[commands.length - 1].requestId;
 
+  if (supportsClusterTime(this.topology) && this.topology.clusterTime) {
+    commands.forEach(command => {
+      if (command instanceof Query) {
+        command.query.$clusterTime = this.topology.clusterTime;
```
kmahar: thought it was supposed to be a top-level field in the command? why is it different for queries?

lib/connection/pool.js
```diff
@@ -31,6 +31,11 @@ var DESTROYED = 'destroyed';
 
 var _id = 0;
 
+function supportsClusterTime(topology) {
```
kmahar: should there be a test checking behavior in the case where cluster time isn't supported?

kmahar:
done commenting

lib/connection/pool.js
mbroadst: all existing mock tests use a lower `maxWireVersion`, so essentially "all other tests" are expressing that code path 😄

lib/connection/pool.js
mbroadst: Good catch. The logic for `resolveClusterTime` used to be here, so this was more relevant.

lib/connection/pool.js
mbroadst: Also good catch.  Originally the spec was designed assuming existence of `OP_MSG` support, which the server is guaranteed to support with `maxWireVersion >= 6`.  Unfortunately, we will not necessarily have that done in time so we need to accommodate `OP_QUERY` _and_ `OP_MSG` forms, and the `OP_QUERY` form expects this field to exist within the top-level `query` field.  I've informed @rstam about this.

kmahar:
cool lgtm

PR #222, opened 2017-09-30
feat(driver-sessions): Implement driver sessions for the core driver

mbroadst:
This is the first step in full support for driver sessions, implementing the `ClientSession` and `ServerSession` types, as well as the `ServerSessionPool`.  The rest of the behavior defined in the spec relates specifically to the porcelain layer ("Where does the pool live?", "Does the topology we're connecting to support sessions?", etc). 

Required reading: https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst

**NOTE:** This branch is based on the `cluster-time-gossiping` branch, so it might look a little big at first. Once #219 is merged this should reduce in size dramatically.

### DIFF
--- a/lib/connection/pool.js
+++ b/lib/connection/pool.js
@@ -1164,6 +1164,9 @@ Pool.prototype.write = function(commands, options, cb) {
       }
 
       sessionOptions.lsid = operation.session.id;
+
+      // update the `lastUse` of the acquired ServerSession
+      operation.session.serverSession.lastUse = Date.now();
     }
 
     // decorate the commands with session-specific details
--- a/lib/connection/pool.js
+++ b/lib/connection/pool.js
@@ -287,7 +287,6 @@ function connectionFailureHandler(self, event) {
     // Flush all work Items on this connection
     while (this.workItems.length > 0) {
       var workItem = this.workItems.shift();
-      // if(workItem.cb) workItem.cb(err);
       if (workItem.cb) workItem.cb(err);
     }
 
--- a/lib/connection/pool.js
+++ b/lib/connection/pool.js
@@ -14,7 +14,8 @@ var inherits = require('util').inherits,
   opcodes = require('../wireprotocol/shared').opcodes,
   compress = require('../wireprotocol/compression').compress,
   compressorIDs = require('../wireprotocol/compression').compressorIDs,
-  uncompressibleCommands = require('../wireprotocol/compression').uncompressibleCommands;
+  uncompressibleCommands = require('../wireprotocol/compression').uncompressibleCommands,
+  resolveClusterTime = require('../topologies/shared').resolveClusterTime;
 
 var MongoCR = require('../auth/mongocr'),
   X509 = require('../auth/x509'),
@@ -546,7 +547,12 @@ function messageHandler(self) {
 
         // Look for clusterTime, and update it if necessary
         if (message.documents[0] && message.documents[0].hasOwnProperty('$clusterTime')) {
-          self.topology.clusterTime = message.documents[0].$clusterTime;
+          const $clusterTime = message.documents[0].$clusterTime;
+          self.topology.clusterTime = $clusterTime;
+
+          if (workItem.session != null) {
+            resolveClusterTime(workItem.session, $clusterTime);
+          }
         }
 
         // Establish if we have an error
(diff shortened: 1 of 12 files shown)


lib/sessions.js
```diff
@@ -0,0 +1,144 @@
...
+    this.lastUse = Date.now();
+  }
+
+  /**
+   *
+   * @param {*} sessionTimeoutMinutes
+   */
+  hasTimedOut(sessionTimeoutMinutes) {
+    const idleTimeMinutes = Math.round(
+      (((Date.now() - this.lastUse) % 86400000) % 3600000) / 60000
```
kmahar: add a comment? not really clear what this formula represents.

lib/sessions.js
```diff
@@ -0,0 +1,144 @@
...
+    while (this.sessions.length) {
+      const session = this.sessions[this.sessions.length - 1];
+      if (session.hasTimedOut(sessionTimeoutMinutes)) {
+        this.sessions.pop();
+      } else {
+        break;
+      }
+    }
+
+    this.sessions.push(session);
```
kmahar: [spec](https://github.com/mongodb/specifications/blob/b1ca52d9b7010ccd40e77d9b5848d33e02a4ab37/source/sessions/driver-sessions.rst#algorithm-to-return-a-serversession-instance-to-the-server-session-pool) seems to say you should only add it to this queue if it's not expiring. I don't think it would have any effect on correctness to add it regardless, since you'll only return it in the future if it's not expired, but it's potentially bad performance-wise?

lib/utils.js
```diff
@@ -31,6 +33,14 @@ var assign = Object.assign
       return to;
     };
 
+const uuidV4 = () => {
```
kmahar: a couple comments might be useful here too - at least for me, bit manipulation code is never really readable on its own. I had to spend some time reading the RFC spec to understand what these ops are doing

lib/connection/pool.js
```diff
@@ -31,7 +32,7 @@ var DESTROYED = 'destroyed';
 
 var _id = 0;
 
-function supportsClusterTime(topology) {
+function hasSessionSupport(topology) {
```
kmahar: this seems to still be implementing the check for whether cluster time is supported. there is a separate description in the spec of how to check if a deployment supports sessions, which involves checking whether logicalSessionTimeoutMinutes is set. I don't see that happening anywhere, is it somehow implicitly checked?

lib/sessions.js
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    // TODO:
+    //   When connected to a sharded cluster the endSessions command
+    //   can be sent to any mongos. When connected to a replica set the
+    //   endSessions command MUST be sent to the primary if the primary
+    //   is available, otherwise it MUST be sent to any available secondary.
+    //   Is it enough to use: ReadPreference.primaryPreferred ?
+
+    this.topology.command(
```
kmahar: this is more of a question about the specs - "drivers MUST ignore any errors returned by the endSessions command". ([here](https://github.com/mongodb/specifications/blob/b1ca52d9b7010ccd40e77d9b5848d33e02a4ab37/source/sessions/driver-sessions.rst#endsessions)) why is that the case? also, is this actually ignoring the error if it passes it to a callback? it seems odd that if the endSessions command fails we still mark the session as hasEnded and release it.

kmahar: also, it seems like this command should have the clusterTime with it if it's being gossiped?

mbroadst: yeah you're correct about that first point, I shouldn't be passing it up the chain.  The `$clusterTime` is added at the pool level, so it will be included with this command as well :)

mbroadst: ✅

lib/sessions.js
mbroadst: ✅

lib/connection/pool.js
mbroadst: The check for `logicalSessionTimeoutMinutes` you reference in the spec will reside in the porcelain layer.  This checks for `$clusterTime` support, but  more importantly checks internally for whether we should look for a provided `session` on the operation - so I renamed it `hasSessionSupport`.

lib/sessions.js
mbroadst: ✅

PR #223, opened 2017-10-03
re issue-220: remove _hiCache, not needed and fixes memory leak

andrasq:
Remove `_hiCache` from `lib/auth/scram.js`.

`_hiCache` was never pruned, and in some environments grew without bound. This leaked memory, used more time to garbage collect, and increased call latencies. Testing showed that removing the cache had no adverse impact compared to mongodb-2.1.18, an older version of the mongo driver without _hiCache.

### DIFF
--- a/lib/auth/scram.js
+++ b/lib/auth/scram.js
@@ -74,18 +74,9 @@ var xor = function(a, b) {
   return new Buffer(res);
 }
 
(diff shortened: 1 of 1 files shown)


mbroadst:
Thanks for your contribution!

PR #225, opened 2017-10-05
cache the ScramSHA1 salted passwords, capping the cache at 200 entries

andrasq:
restores the`_hiCache` functionality, but without the unbounded cache growth


mbroadst:
@andrasq thanks for the contribution!

PR #229, opened 2017-10-12
fix(auth-scram): fixing an issue where the unsalted password is being cached

arosequist:
With 2.1.16, only the first auth will succeed, because subsequent calls with the same password will return the unsalted one.


mbroadst:
@arosequist thanks for the PR, I merged a fix without seeing your PR!

arosequist:
Thanks for pushing a new version of this and mongodb/node-mongodb-native so quickly!

mbroadst:
Many apologies for the breakage!  Very focused on new 3.6 features at the moment, so this unfortunately slipped through the cracks 😞

PR #230, opened 2017-10-31
fix(connection): default `family` to undefined rather than 4

vkarpov15:
Upon closer inspection, #183 was actually very much backwards breaking and should not have been merged. This issue causes https://github.com/Automattic/mongoose/issues/5760 via https://github.com/Automattic/mongoose/commit/93e771ce5678c20da3a8dd08a4b77de8a1380446 .

The problem with #183 is that [`net.createConnection()`](https://nodejs.org/api/net.html) calls [`dns.lookup()`](https://nodejs.org/api/dns.html#dns_dns_lookup_hostname_options_callback) with the `family` option always set. If the `family` option is not set, `dns.lookup()` will return both ipv4 and ipv6 addresses, but since mongodb-core always sets `family` now, the driver can only connect to ipv4 or ipv6 depending on the value of `family`. In particular, this means that any user that was connecting to an ipv6 address now **must** specify `family: 6` in porcelain layer >= 2.2.28 and mongoose >= 4.11.6 in order to successfully connect, whereas they didn't have to before.

### DIFF
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -23,7 +23,7 @@ var connections = {};
  * @class
  * @param {string} options.host The server host
  * @param {number} options.port The server port
- * @param {number} [options.family=4] Version of IP stack. Defaults to 4.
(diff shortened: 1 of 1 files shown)


mbroadst:
@jlord this is needs to be ported to 3.x as well please

PR #235, opened 2017-11-14
fix(connection): fixing leak in 3.0.0

daprahamian:
Adding the fix from #234 to 3.0.0 branch.

### DIFF
--- a/test/tests/functional/pool_tests.js
+++ b/test/tests/functional/pool_tests.js
@@ -39,6 +39,53 @@ describe('Pool tests', function() {
     }
   });
 
+  it.only('Should only listen on connect once', {
+    metadata: { requires: { topology: 'single' } },
+
+    test: function(done) {
+      // Enable connections accounting
+      Connection.enableConnectionAccounting();
+
+      // Attempt to connect
+      var pool = new Pool(null, {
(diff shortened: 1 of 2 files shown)


test/tests/functional/pool_tests.js
```diff
@@ -39,6 +39,36 @@ describe('Pool tests', function() {
...
+        messageHandler: function() {}
+      });
+
+      // Add event listeners
+      pool.on('connect', function(_pool) {
+        var connections = _pool.allConnections();
+        expect(connections).to.have.lengthOf(1).and.to.have;
+        expect(connections[0].connection.listenerCount('connect')).to.equal(1);
+        _pool.destroy();
+        Connection.disableConnectionAccounting();
```
mbroadst: I think what we actually want to test is that there is a single listener count until `connect` occurs right? So there should be an initial listener count for that event, and when connect occurs there should be zero listeners.  Or am I misunderstanding something?

daprahamian: I'd expect the listener count to be 0 as well, but there is some listener nestled deep inside of node's core that we don't have control over.

Still, I will refactor to emphasize that the number of listeners for the connection should decrease by 1.

PR #236, opened 2017-11-15
refactor(cluster-time): topologies should report consistent times

mbroadst:
The ReplSet and Mongos topologies are really just containers for a
number of Server topologies.  Hence, when initially implementing
cluster time gossipping a `clusterTimeWatcher` was used to
coordinate reported times from child topologies. Unfortunately,
this ignored the need to _read_ the deployment (parent) cluster
time.  Instead Server topologies internal to the container
topologies now have a concept of a `parent`, which also simplifies
the code dramatically.

NODE-1088

**NOTE:** this also contains an additional fix for temporarily skipping the 1:1 topology guarantee of the driver sessions spec, please disregard during review

### DIFF
--- a/test/tests/unit/single/sessions_tests.js
+++ b/test/tests/unit/single/sessions_tests.js
@@ -544,7 +544,7 @@ describe('Sessions (Single)', function() {
     }
   });
 
-  it('should not allow use of session object across clients', {
+  it.skip('should not allow use of session object across clients', {
     metadata: { requires: { topology: 'single' } },
     test: function(done) {
       const client = new Server(test.server.address());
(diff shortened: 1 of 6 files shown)


test/tests/unit/single/sessions_tests.js
```diff
@@ -544,7 +544,7 @@ describe('Sessions (Single)', function() {
     }
   });
 
-  it('should not allow use of session object across clients', {
+  it.skip('should not allow use of session object across clients', {
```
daprahamian: Is this disabled permanently?

mbroadst: no, but there's limited time to solve it right now, and this requirement is far less important in node than in multithreaded environments

PR #237, opened 2017-11-16
feat(replset): more verbose replica set errors emission 

daprahamian:
Port of #231 to 3.0.0.
When a replica set topology has an invalid name, core emits an error
stating that no primary or secondary found in the replica set, which can
cause some user confusion. This simply expands the language to include
that the name may be invalid to help with debugging.


mbroadst:
@daprahamian can you change the actual commit message to be more informative, this will be aggregated into a release changelog

daprahamian:
Commit message updated

PR #240, opened 2017-11-20
[Ported] fix(connection): default `family` to undefined rather than 4

jlord:
This ports #230 fix to the 3.0 branch.


lib/connection/connection.js
```diff
@@ -542,9 +542,15 @@ Connection.prototype.connect = function(_options) {
...
   // Create new connection instance
-  var connection_options = self.domainSocket
-    ? { path: self.host }
-    : { port: self.port, host: self.host, family: self.family };
+  var connection_options;
+  if (self.domainSocket) {
+    connection_options = { path: self.host };
+  } else {
+    connection_options = { port: self.port, host: self.host };
+    if (self.fmaily !== void 0) {
```
daprahamian: family

jlord: 😱

Issue #242, opened 2017-11-23
Missed ssl ecdhCurve option


newsiberian:
Hi, there is a new ssl option [was recently added](https://github.com/nodejs/node/pull/15206) to node.js.

W/o it support in `mongodb-core` it is impossible to set connection with mongo if it is used certificate with ECDH curve.

Here is related [issue](https://github.com/nodejs/node/issues/16196)

PR #243, opened 2017-11-23
fix(connection): added missed tls option ecdhCurve

newsiberian:
Hi, this is related with #242.


newsiberian:
This PR not provide a support of `ecdhCurve` to usage in combine with `mongo-native-driver`. It should be done separately. This PR allows to establish connection from `mongodb-core` itself

Issue #247, opened 2017-11-23
ReplSet.topologyMonitor seem to be causing memory leaks via Timeout uncontrolled instantiation


ilguzin:
In my recent investigation of node.js application memory leaks the heapdump files comparison revealed enormous linear increase in amount of Timeout objects under the ReplSet objects.
Googling has gave me this: https://jira.mongodb.org/browse/NODE-987
Despite the ticket has been resolved the issue seem to be still existing.

In the solution in the related ticket (NODE-987) there was a filtering applied to completed Timeouts. Later it was improved via Timeout.isRunning method. But the deal is that in replset.js::topologyMonitor -> executeReconnect function does not "stops" (Timeout.stop()) the Timeout instance after it has finished and they still continue to utilise the memory.

Please fix that, unless i'm missing something :-) . Thanks.

PR #248, opened 2017-11-24
Artificial _called property set on Timeout instance...

ilguzin:
... to have Timeout.isRunning() work properly for legacy node.js versions. Resolves mongodb-js/mongodb-core#247.

Fixing Timeout.isRunning was probably the best way to solve the issue and fit into the initial idea behind Timeout class.

Also returned back 0.10 and 0.12 targets for Travis. For that reason the 'request' lib has been fixed to @2.65.0 version in dev deps and added tests for the memory leak issue.

It does not seem to be reasonable to have these legacy targets to be turned off simply because of Hawk semver. That move has eventually played a dramatic role in hidding of the Timeout._called absence for v0.12 which was causing memory leak.

### DIFF
--- a/lib/topologies/shared.js
+++ b/lib/topologies/shared.js
@@ -217,7 +217,8 @@ function Timeout(fn, time) {
 
   this.start = function () {
     if (!this.isRunning()) {
-      timer = setTimeout(function() { fn(); if (timer !== false) { timer._called = true; } }, time);
+      // The artificial _called is set here for compatibility with node.js 0.10.x/0.12.x versions
+      timer = setTimeout(function() { fn(); if (timer !== false && timer._called === undefined) { timer._called = true; } }, time);
     }
     return this;
   };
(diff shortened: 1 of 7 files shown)


mbroadst:
@daprahamian please take a look at this

lib/topologies/shared.js
```diff
@@ -217,7 +217,8 @@ function Timeout(fn, time) {
 
   this.start = function () {
     if (!this.isRunning()) {
-      timer = setTimeout(fn, time);
+      // The artificial _called is set here for compatibility with node.js 0.10.x/0.12.x versions
+      timer = setTimeout(function() { fn(); if (timer !== false && timer._called === undefined) { timer._called = true; } }, time);
```
daprahamian: Hi @ilguzin,

Can you please incorporate the following changes:

1. replace `timer !== false` with just `timer`
2. Break this all out into multiple lines.

Ideally, it would be something like this?

```js
timer = setTimeout(function() {
  fn();
  if (timer && timer._called === undefined) {
    timer._called = true;
  }
}, time);
```

ilguzin:
Any chance to see it in next release this month?

mbroadst:
@ilguzin there will be a 2.x release next week

ilguzin:
@mbroadst, cool! Thanks :-)

PR #250, opened 2017-11-29
feat(retryable-writes): add basic support for retryable writes to mongodb-core

mbroadst:
ticket: https://jira.mongodb.org/browse/NODE-1105
spec: https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst

This implements the basic building blocks for support for retryable writes in the core driver.  The code sections are implementations of the pseudocode documented in the spec linked above.  The feature is "enabled" on supported servers by passing a `retryWrites: true` option down to the topology's supported write operations (insert/update/remove), as well as a valid session.  All other details should be covered by the spec.

It should be mentioned that this includes a preliminary bug fix for the pool supporting a "minimum pool size".  This was a problem encountered during testing because the simulation of a lost connection resulted in a pool which never increased in size for a subsequent write operation - not a situation that's likely to occur in the real world, but something we eventually want in a pool refactor.  It defaults to 0 (no change in behavior from the status quo), but works well enough for the test case I needed.  Consider this work incomplete, but working.

### DIFF
--- a/lib/topologies/mongos.js
+++ b/lib/topologies/mongos.js
@@ -15,7 +15,7 @@ const inherits = require('util').inherits,
   createClientInfo = require('./shared').createClientInfo,
   SessionMixins = require('./shared').SessionMixins,
   isRetryableWritesSupported = require('./shared').isRetryableWritesSupported,
-  txnNumber = require('./shared').txnNumber;
+  getNextTransactionNumber = require('./shared').getNextTransactionNumber;
 
 const BSON = retrieveBSON();
 
@@ -896,7 +896,7 @@ var executeWriteOperation = function(self, op, ns, ops, options, callback) {
   }
 
   // increment and assign txnNumber
-  options.txnNumber = txnNumber(options.session);
+  options.txnNumber = getNextTransactionNumber(options.session);
 
   server[op](ns, ops, options, (err, result) => {
     if (!err) return callback(null, result);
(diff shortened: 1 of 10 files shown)


lib/topologies/shared.js
```diff
@@ -402,6 +405,36 @@ const SessionMixins = {
...
+
+  return true;
+};
+
+/**
+ * Increment the transaction number on the ServerSession contained by the provided ClientSession
+ *
+ * @param {ClientSession} session
+ */
+const txnNumber = function(session) {
```
daprahamian: Do we need to abbreviate this to `txnNumber`? Can we name it `transactionNumber`, or better `getNextTransactionNumber`?

lib/topologies/mongos.js
```diff
@@ -879,18 +882,39 @@ Mongos.prototype.isDestroyed = function() {
...
+
   // Pick a server
-  var server = pickProxy(self);
+  let server = pickProxy(self);
   // No server found error out
   if (!server) return callback(new MongoError('no mongos proxy available'));
-  // Execute the command
-  server[op](ns, ops, options, callback);
+
+  if (!options.retryWrites || !options.session || !isRetryableWritesSupported(self)) {
```
daprahamian: Should we DeMorgan this?

mbroadst: I know its slightly harder to read this way, but imho it actually makes the code much more readable - basically what's being done here is "if none of this is set, then run this minimal piece of code and exit out, otherwise do this much longer/more involved segment."  It also reduces the brace level

PR #251, opened 2017-11-30
Don't redeclare BSON variable

denizdogan:
This fixes a problem that arises when using Rollup to bundle an application that uses this library.

```
[!] Error: Identifier 'BSON' has already been declared
node_modules/mongodb-core/lib/auth/plain.js (11:4)
 9:   , MongoError = require('../error');
10:
11: var BSON = retrieveBSON();
        ^
12:
13: var AuthSession = function(db, username, password) {
```

### DIFF
--- a/lib/auth/plain.js
+++ b/lib/auth/plain.js
@@ -8,7 +8,7 @@ var f = require('util').format
   , Query = require('../connection/commands').Query
   , MongoError = require('../error');
 
-var BSON = retrieveBSON();
+BSON = retrieveBSON();
 
 var AuthSession = function(db, username, password) {
   this.db = db;
--- a/lib/auth/plain.js
+++ b/lib/auth/plain.js
@@ -3,12 +3,12 @@
 var BSON = require('bson');
 
 var f = require('util').format
-  , Binary = BSON.Binary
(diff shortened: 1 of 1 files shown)


denizdogan:
If anyone else has this problem and this PR doesn't get merged, I solved my problems temporarily by using `rollup-plugin-replace` as such:

```js
import replace from 'rollup-plugin-replace'

export default {
  plugins: [
    ...
    replace({
      delimiters: ['', ''],
      'var BSON = retrieveBSON();': 'BSON = retrieveBSON();'
    }),
    ...
  ]
}
```

mbroadst:
@denizdogan Hi, thanks for this!  I actually think the code block should look like this:

```
var f = require('util').format
  , retrieveBSON = require('../connection/utils').retrieveBSON
  , Query = require('../connection/commands').Query
  , MongoError = require('../error');

var BSON = retrieveBSON()
  , Binary = BSON.Binary;
```

denizdogan:
@mbroadst Thanks for the swift response, much appreciated! I've updated my branch accordingly :)

mbroadst:
@denizdogan looks good, thanks for your contribution!

PR #253, opened 2017-12-04
refactor(read-preference): merge missing api from native version

mbroadst:
This updates the ReadPreference in the core module to serve as a
full replacement for the one in the porcelain layer.

NODE-1187

### DIFF
--- a/lib/topologies/read_preference.js
+++ b/lib/topologies/read_preference.js
@@ -63,6 +63,17 @@ Object.defineProperty(ReadPreference.prototype, 'preference', {
   }
 });
 
+const VALID_MODES = [
+  ReadPreference.PRIMARY,
+  ReadPreference.PRIMARY_PREFERRED,
+  ReadPreference.SECONDARY,
+  ReadPreference.SECONDARY_PREFERRED,
+  ReadPreference.NEAREST,
+  true,
+  false,
+  null
+];
+
 /**
  * Validate if a mode is legal
  *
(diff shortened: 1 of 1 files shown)


lib/topologies/read_preference.js
```diff
@@ -54,13 +55,54 @@ var ReadPreference = function(preference, tags, options) {
...
+});
+
+/**
+ * Validate if a mode is legal
+ *
+ * @method
+ * @param {string} mode The string representing the read preference mode.
+ * @return {boolean}
+ */
+ReadPreference.isValid = function(mode) {
```
daprahamian: Can we use a `Set` here?

lib/topologies/read_preference.js
```diff
@@ -54,13 +55,54 @@ var ReadPreference = function(preference, tags, options) {
   }
 };
 
+// Support the deprecated `preference` property introduced in the porcelain layer
+Object.defineProperty(ReadPreference.prototype, 'preference', {
```
daprahamian: In the future we should replace this whole thing with a class and a getter here

mbroadst: agreed, but a bit pressed for time today :)

lib/topologies/read_preference.js
mbroadst: like:
```
const validModes = [
  ReadPreference.PRIMARY, 
  ReadPreference.PRIMARY_PREFERRED,
  ReadPreference.SECONDARY,
  ReadPreference.SECONDARY_PREFERRED,
  ReadPreference.NEAREST,
  true,
  false,
  null
];

ReadPreference.isValid = function(mode) {
  return validModes.has(mode);
}
```
?

daprahamian: Yeah, either by using the native `Set` (do we get that in Node 4?), or by just doing `validModes.indexOf(x) >=0`

mbroadst: okay I'm into it

PR #258, opened 2017-12-14
fix(secondaries): fixes connection with secondary readPreference

daprahamian:
Ensures that when readPreference is `secondary`, the `connect`
event is not triggered until connection has been established
with a secondary.

Fixes NODE-1089

### DIFF
--- a/test/tests/unit/replset/read_preference_tests.js
+++ b/test/tests/unit/replset/read_preference_tests.js
@@ -1,0 +1,51 @@
+'use strict';
+
+const expect = require('chai').expect;
+const ReplSet = require('../../../../lib/topologies/replset');
+const ReadPreference = require('../../../../lib/topologies/read_preference');
+const mock = require('../../../mock');
+const ReplSetFixture = require('../common').ReplSetFixture;
+
+describe('Secondaries (ReplSet)', function() {
+  let test;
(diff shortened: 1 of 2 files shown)


test/tests/functional/rs_mocks/read_preferences_tests.js
```diff
@@ -1337,4 +1337,104 @@ describe('ReplSet Read Preferences (mocks)', function() {
       });
     }
   });
+
```
mbroadst: Can we please move this test to `tests/unit`, and use the `ReplSetFixture` to cut out a lot of the code dupe, and allow us to focus on the actual test conditions.

test/tests/unit/replset/read_preference_tests.js
```diff
@@ -0,0 +1,51 @@
...
+          haInterval: 100,
+          size: 1
+        }
+      );
+
+      replSet.on('error', done);
+
+      replSet.on('connect', server => {
+        let err;
+        try {
```
mbroadst: do we actually need this try/catch? if the assertion fails we will still get the message during the test run (e.g. I think there is already a try/catch internal to mocha that provides this same behavior)

daprahamian: since it is asynchronous, it won't work

PR #260, opened 2017-12-18
feat(connection): attempt both ipv6 and ipv4 when no family entered

daprahamian:
If the user does not specify an IP family, we will attempt to
connect first via IPv6, and fall back to IPv4 if that fails

Currently WIP as I need to write unit tests for this.

### DIFF
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -528,100 +528,131 @@ function merge(options1, options2) {
   }
 }
 
+function makeSSLConnection(self, _options) {
+  let sslOptions = {
+    socket: self.connection,
+    rejectUnauthorized: self.rejectUnauthorized
+  };
+
+  // Merge in options
+  merge(sslOptions, this.options);
+  merge(sslOptions, _options);
+
+  // Set options for ssl
+  if (self.ca) sslOptions.ca = self.ca;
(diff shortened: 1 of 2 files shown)


daprahamian:
Removed WIP, this should be ready for review

test/tests/functional/connection_tests.js
```diff
@@ -0,0 +1,127 @@
+'use strict';
+
+var bson = require('bson');
```
jlord: A lone `var`

lib/connection/connection.js
```diff
@@ -528,100 +528,131 @@ function merge(options1, options2) {
...
-  _options = _options || {};
-  // Set the connections
-  if (connectionAccounting) addConnection(this.id, this);
-  // Check if we are overriding the promoteLongs
-  if (typeof _options.promoteLongs === 'boolean') {
-    self.responseOptions.promoteLongs = _options.promoteLongs;
-    self.responseOptions.promoteValues = _options.promoteValues;
-    self.responseOptions.promoteBuffers = _options.promoteBuffers;
+function makeSSLConnection(self, _options) {
+  var sslOptions = {
```
jlord: Lone `var`

daprahamian:
Updated to remove `var`s

PR #263, opened 2017-12-21
fix(sdam): more explicit wire protocol error message

jlord:
This updates the error message received when a user attempts to connect to a server running an older version of MongoDB than the latest Node.js Driver version supports. For instance, using the 3.x version of the driver to connect to a server running something below MongoDB 2.6. 

@daprahamian and I paired on this PR 💯 

@mbroadst do you think we should break up the error messages into other files to do more explicit testing?

NODE-1186

### DIFF
--- a/test/tests/functional/server_tests.js
+++ b/test/tests/functional/server_tests.js
@@ -1055,7 +1055,7 @@ describe('Server tests', function() {
 
       test: function(done) {
         server.setMessageHandler(request => {
-          request.reply({ ok: 1, maxWireVersion: 1 });
+          request.reply(Object.assign({}, mock.DEFAULT_ISMASTER, { maxWireVersion: 1 }));
         });
 
         const client = new Server(server.address());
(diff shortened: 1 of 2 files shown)


test/tests/functional/server_tests.js
```diff
@@ -1043,4 +1044,36 @@ describe('Server tests', function() {
...
+    let server;
+    beforeEach(() => mock.createServer().then(_server => (server = _server)));
+    afterEach(() => mock.cleanup());
+
+    it('errors when unsupported wire protocol is returned from isMaster', {
+      metadata: { requires: { topology: ['single'] } },
+
+      test: function(done) {
+        server.setMessageHandler(request => {
+          request.reply({ ok: 1, maxWireVersion: 1 });
```
mbroadst:
```
request.reply(Object.assign({}, mock.DEFAULT_ISMASTER, { maxWireVersion: 1 }));
```

PR #270, opened 2018-01-26
Cannot set _called property on number.

durran:
Background:

Compass is currently generating tons of errors when connecting to replica sets on our latest master, because of the `Timeout` object in the shared topology:

screen shot 2018-01-26 at 12 49 22 pm

Basically it's trying to set a property `_called` on a number, which is invalid in this context for some reason (I can do this from the console all day long with no issue). The number is the return value of the `setTimeout` function, which is the expected return value of that function.

There was a comment that the `_called` property was there for Node compatibility, altough I'm not sure why it was needed. First of all, Node's `Timeout` object has this set as an instance property on the `Timeout` object itself, not the result of the `setTimeout` method, so the code wouldn't be API compatible with that object anyways.

https://github.com/nodejs/node/blob/v7.9.0/lib/timers.js#L478

I haven't been able to peg down exactly why this started happening or exactly when as of yet, but I have tested this new code in Compass and the error goes away.

### DIFF
--- a/lib/topologies/shared.js
+++ b/lib/topologies/shared.js
@@ -214,14 +214,14 @@ function Interval(fn, time) {
 
 function Timeout(fn, time) {
   var timer = false;
+  this._called = false;
 
   this.start = function () {
     if (!this.isRunning()) {
       timer = setTimeout(function() {
         fn();
-        if (timer && timer._called === undefined) {
-          // The artificial _called is set here for compatibility with node.js 0.10.x/0.12.x versions
-          timer._called = true;
+        if (timer && !this._called) {
+          this._called = true;
         }
       }, time);
     }
@@ -235,7 +235,7 @@ function Timeout(fn, time) {
   };
 
   this.isRunning = function () {
-    if(timer && timer._called) return false;
+    if (timer && this._called) return false;
     return timer !== false;
   };
 }


mbroadst:
@durran thanks for the report. This was changed in #247 to alleviate some memory leaks in older versions of node. v0.12, for instance, [does not set this value](https://github.com/nodejs/node/blob/v0.12.6-release/lib/timers.js#L296-L303), and I wonder why later versions of node are having an issue though? 

The error you're receiving indicates to me that the result of `setTimeout` is no longer a `Timeout` instance, but instead a number - is that possible? The docs still show that `Timeout`s are returned.

durran:
In node it’s a Timeout object, but in the browser it’s a number.

mbroadst:
@durran ah good point, sorry wasn't thinking electron. I'll give this another look over after lunch

durran:
@mbroadst https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout

My changes would work both in Node and in browser. Although I still have issues running this test suite on my machine. It’s always failing.

mbroadst:
@durran yeah the 2.0 test suite is difficult to use, the 3.0.0 branch is current and using mocha wrapped with the mongodb-test-runner

dickeylth:
Could the `2.x` branch be patched? I've got a project which is still depending on the `2.x` version of mongodb.

PR #277, opened 2018-02-13
Help 5834:  fix(sessions): actually allow ending of sessions

daprahamian:
Currently, endSessions filters out any calls that happen while
the topology is connected. However, it is impossible to send
an endSessions command if you are not connected. This fix allows
sessions to be ended while the topology is connected.

Also adds a helper method that `native` can call to easily end all pooled sessions.

WIP b/c this currently breaks APM tests in native.

### DIFF
--- a/lib/sessions.js
+++ b/lib/sessions.js
@@ -141,6 +141,13 @@ class ServerSessionPool {
     this.sessions = [];
   }
 
+  endAllPooledSessions() {
+    if (this.sessions.length) {
+      this.topology.endSessions(this.sessions.map(session => session.id));
+      this.sessions = [];
+    }
+  }
+
   /**
    * @returns {ServerSession}
    */
--- a/lib/topologies/shared.js
+++ b/lib/topologies/shared.js
@@ -378,11 +378,6 @@ function resolveClusterTime(topology, $clusterTime) {
 //       to share code.
 const SessionMixins = {
   endSessions: function(sessions, callback) {
-    if (this.isConnected()) {
-      if (typeof callback === 'function') callback();
-      return;
-    }
-
     if (!Array.isArray(sessions)) {
       sessions = [sessions];
     }
@@ -395,7 +390,7 @@ const SessionMixins = {
     //   Is it enough to use: ReadPreference.primaryPreferred ?
     this.command(
       'admin.$cmd',
-      { endSessions: sessions.map(s => s.id) },
+      { endSessions: sessions },
       { readPreference: ReadPreference.primaryPreferred },
       () => {
         // intentionally ignored, per spec


daprahamian:
This should be ready for review @mbroadst @jlord

lib/sessions.js
```diff
@@ -138,16 +138,42 @@ class ServerSessionPool {
     }
 
     this.topology = topology;
-    this.sessions = [];
+    this.pool = [];
```
mbroadst: I'm not sure `pool` is a more descriptive name, it certainly isn't in the unit tests lower in the file where we are checking `pool.pool` (also we could avoid about half the changes in this PR if we just leave this as `sessions` 😄)

lib/topologies/shared.js
```diff
@@ -395,7 +390,7 @@ const SessionMixins = {
     //   Is it enough to use: ReadPreference.primaryPreferred ?
     this.command(
       'admin.$cmd',
-      { endSessions: sessions.map(s => s.id) },
+      { endSessions: sessions },
```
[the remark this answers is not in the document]
mbroadst: also it appears that either this code path is wrong, or [this](https://github.com/mongodb-js/mongodb-core/pull/277/files#diff-a71253d0ae71ba8f0fc759adbd6ba543R165) one is.

lib/sessions.js
```diff
@@ -162,17 +188,17 @@ class ServerSessionPool {
...
-      const session = this.sessions[this.sessions.length - 1];
+    while (this.pool.length) {
+      const session = this.pool[this.pool.length - 1];
       if (session.hasTimedOut(sessionTimeoutMinutes)) {
-        this.sessions.pop();
+        this.pool.pop();
       } else {
         break;
       }
     }
```
mbroadst: I think you're forgetting to remove the session from `active` in this case? Perhaps you're depending on the emitted `ended` event?

daprahamian: yeah, im relying on the `ended` event

lib/topologies/shared.js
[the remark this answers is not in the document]
daprahamian: There are currently cases where [we call endSessions with just the sessionId](https://github.com/mongodb-js/mongodb-core/blob/3.0.0/lib/sessions.js#L58), and cases where [we instead have access to the entire session](https://github.com/mongodb/node-mongodb-native/pull/1665/commits/1f1c7c3e45b319410dafa2dfbf29b1f145e7c3f8#diff-d3c4c7ddb4c090ab1d114107933a2ce0R401). To simplify, I am removing the map function from here, and instead using it where it is necessary.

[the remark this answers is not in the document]
mbroadst: oh I see, you're just moving the mapping up a level - fine

lib/topologies/shared.js
```diff
@@ -378,11 +378,6 @@ function resolveClusterTime(topology, $clusterTime) {
 //       to share code.
 const SessionMixins = {
   endSessions: function(sessions, callback) {
-    if (this.isConnected()) {
```
mbroadst: this is insane, I can't believe I committed it 😄 It _must_ have been accidentally inverted? `if (!this.isConnected()) { // bail }`?

PR #280, opened 2018-03-02
feat(sessions): adding implicit cursor session support

daprahamian:
Adds ability to tell difference between implicit and explicit
sessions, and adds part of implicit session support to
cursors.

Part of NODE-1326

### DIFF
--- a/lib/cursor.js
+++ b/lib/cursor.js
@@ -171,6 +171,27 @@ Cursor.prototype.cursorSkip = function() {
   return this.cursorState.skip;
 };
 
+Cursor.prototype._endSession = function(options, callback) {
+  if (typeof options === 'function') {
+    callback = options;
+    options = {};
+  }
+  options = options || {};
+
+  const session = this.cursorState.session;
+
+  if (session && (options.force || session.owner === this)) {
+    this.cursorState.session = undefined;
+    session.endSession(callback);
+    return true;
+  }
+
+  if (callback) {
+    callback();
+  }
+  return false;
+};
+
 //
 // Handle callback (including any exceptions thrown)
 var handleCallback = function(callback, err, result) {
@@ -475,15 +496,11 @@ Cursor.prototype.rewind = function() {
  */
 var isConnectionDead = function(self, callback) {
   if (self.pool && self.pool.isDestroyed()) {
-    self.cursorState.notified = true;
     self.cursorState.killed = true;
-    self.cursorState.documents = [];
-    self.cursorState.cursorIndex = 0;
-    callback(
-      new MongoNetworkError(
-        f('connection to host %s:%s was destroyed', self.pool.host, self.pool.port)
-      )
+    const err = new MongoNetworkError(
+      f('connection to host %s:%s was destroyed', self.pool.host, self.pool.port)
     );
+    _setCursorNotifiedImpl(self, () => callback(err));
     return true;
   }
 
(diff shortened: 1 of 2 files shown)


lib/sessions.js
```diff
@@ -39,6 +39,12 @@ class ClientSession extends EventEmitter {
     }
 
     this.operationTime = null;
+
+    this.explicit = !!options.explicit;
+  }
+
+  isExplicit() {
```
mbroadst: tbh I'm inclined to say we can just access the `.explicit` property explicitly, pun intended. Otherwise this is at very least a `get` accessor

lib/cursor.js
```diff
@@ -654,6 +669,10 @@ var nextFunction = function(self, callback) {
     self._find(function(err) {
       if (err) return handleCallback(callback, err, null);
 
+      if (self.cursorState.cursorId && self.cursorState.cursorId.isZero() && self._endSession) {
```
mbroadst: where you running into cases where `self_endSession` didn't exist?

lib/cursor.js
```diff
@@ -171,6 +171,27 @@ Cursor.prototype.cursorSkip = function() {
   return this.cursorState.skip;
 };
 
+Cursor.prototype._endSession = function(options, callback) {
```
mbroadst: this is more of a nitpick/question than a mandate, but for private methods like this I generally keep them _off_ the prototype unless I explicitly need it for testing. I would opt for a:
```
function cursorEndSession(cursor, options, callback)
```
that isn't exported by the file.

Is there a reason you prefer your approach here? Do you anticipate people will need to override this behavior, or that we will need to futz with it in testing?

lib/cursor.js
```diff
@@ -171,6 +171,27 @@ Cursor.prototype.cursorSkip = function() {
...
+    options = {};
+  }
+  options = options || {};
+
+  const session = this.cursorState.session;
+
+  if (session && (options.force || !session.isExplicit())) {
+    this.cursorState.session = undefined;
+    session.endSession(callback);
+    return true;
```
mbroadst: why does this method return a boolean? Actually, this might be a good time to bring up the question: why isn't this method documented at all 😄

lib/sessions.js
daprahamian: Ok, that is doable

lib/cursor.js
daprahamian: Yes, this is a bigger issue I had to work around: [commandCursor selectively inherits methods from the base class](https://github.com/mongodb/node-mongodb-native/blob/3.0.0/lib/command_cursor.js#L128). I'm not sure why this is the case, and I wasn't sure if command cursors should have the same session management code (especially b/c when I tried to add it in, it led to >200 unit test failures :) ).

lib/cursor.js
daprahamian: The method is overridden in the native PR [here](https://github.com/mongodb/node-mongodb-native/pull/1668/files#diff-14b597bc6604e1b7ec69ae786619e979R217). I'm not particularly happy about this though.

lib/cursor.js
daprahamian: As with the above comment, the boolean value is used [here](https://github.com/mongodb/node-mongodb-native/pull/1668/files#diff-14b597bc6604e1b7ec69ae786619e979R217). As for documentation, I consider it a "private" method. But I can still add a `/** @private */` or `/** @ignore */`.

PR #286, opened 2018-03-21
WIP: Initial pass at prefetching the next 'getmore' call

herbrandson:
This PR is a work-in-progress and is not intended to be merged in at this point. It addresses this feature request: https://jira.mongodb.org/browse/NODE-1375

The idea behind the feature is to not wait for the processing of the current batch of documents to be completed _before_ making the next `getmore` call. Instead, the idea is to better pipeline the process by making the next `getmore` call _in parallel_ with the processing of the current batch.

This is critical for our application as we have found that we spend a significant amount of time waiting on the next batch of results to process.

This PR is only intended to:
1. Show a possible implementation
2. See if there is any interest from the maintainers of mongodb-core in such a PR
3. Determine if this is a reasonable approach to the proposed feature
4. See if there are any implementation details that I'm not aware of that would need to be addressed for this to work

Thanks in advance for your feedback :)

### DIFF
--- a/lib/cursor.js
+++ b/lib/cursor.js
@@ -576,6 +576,25 @@ var clearCursorDocuments = function(self) {
   self.cursorState.cursorIndex = 0;
 };
 
+var dequeueNextBatch = function(self) {
+  var nextBatch = self.cursorState.nextBatches.shift();
+  self.cursorState.documents = nextBatch.documents;
+};
+
+var prefetch = function(self) {
+  // If prefetching has been enabled, request the next batch now
+  if (!self.options.prefetchBatches) return;
+
+  getMore(self, function(err) {
+    if (err) {
+      self.cursorState.prefetchingError = err;
+      if (self.logger.isInfo()) {
+        self.logger.info('An error occurred prefetching next batch.', err);
+      }
+    }
+  });
+};
+
 var getMore = function(self, callback) {
   // If the next batch has already been requested, exit early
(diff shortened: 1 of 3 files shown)


lib/wireprotocol/2_6_support.js
```diff
@@ -140,8 +140,10 @@ WireProtocol.prototype.getMore = function(
     var cursorId = typeof r.cursorId === 'number' ? Long.fromNumber(r.cursorId) : r.cursorId;
 
     // Set all the values
-    cursorState.documents = r.documents;
-    cursorState.cursorId = cursorId;
+    cursorState.nextBatches.push({
```
herbrandson: Rather then setting `cursorState.documents` directly, we store the results in a queu

lib/cursor.js
```diff
@@ -358,7 +360,7 @@ Cursor.prototype._killcursor = function(callback) {
   this.cursorState.dead = true;
   this.cursorState.killed = true;
   // Remove documents
-  this.cursorState.documents = [];
+  clearCursorDocuments(this);
```
herbrandson: There is one slight difference between the original code and what I have hear in that `clearCursorDocuments` will reset both `cursorState.documents` AND `cursorState.cursorIndex`. The original code did not reset the `cursorIndex`.  This seems like a trivial difference, but I thought I'd point it out.

PR #287, opened 2018-03-26
Initial work to test mongodb-core on evergreen

mbroadst:
This includes a number of changes, but generally provides the foundation for testing our two drivers in evergreen, closing: NODE-871, NODE-873, NODE-879, NODE-880, NODE-881, NODE-883, NODE-884, NODE-1191.  It also includes a few modifications to the actual tests to reduce flakiness of tests. 

You can see it all builds in evergreen here: https://evergreen.mongodb.com/version/5ab7939ce3c331475bc1144a. The red boxes in the result are related to node-snappy not being built to support big endian on those ZAP systems, I have a fix for that [here](https://github.com/kesla/node-snappy/pull/152)


test/tests/functional/basic_single_server_auth_tests.js
```diff
@@ -8,7 +8,8 @@ var expect = require('chai').expect,
   Bson = require('bson');
 
 describe('Basic single server auth tests', function() {
```
daprahamian: Is every test `skip`ed in this `describe` block? We could just do `describe.skip`

test/config.js
```diff
@@ -2,6 +2,11 @@
 const ConfigurationBase = require('mongodb-test-runner').ConfigurationBase;
 const f = require('util').format;
 
+const chai = require('chai');
```
daprahamian: Is there a reason we are adding these things in?

lib/connection/pool.js
```diff
@@ -210,6 +210,7 @@ function stateTransition(self, newState) {
   // Get current state
   var legalStates = legalTransitions[self.state];
   if (legalStates && legalStates.indexOf(newState) !== -1) {
+    self.emit('stateChanged', self.state, newState);
```
daprahamian: Is this a bug that needs to be fixed?

.evergreen/run-tests.sh
```diff
@@ -0,0 +1,49 @@
...
+get_distro
+if [ -z "$MONGODB_DOWNLOAD_URL" ]; then
+    get_mongodb_download_url_for "$DISTRO" "$MONGODB_VERSION"
+fi
+# Even though we have the MONGODB_DOWNLOAD_URL, we still call this to get the proper EXTRACT variable
+get_mongodb_download_url_for "$DISTRO"
+download_and_extract "$MONGODB_DOWNLOAD_URL" "$EXTRACT"
+
+# run tests
+OS=$(uname -s | tr '[:upper:]' '[:lower:]')
```
daprahamian: I feel that things like this could be made into bash functions to make it clearer what is happening.

.evergreen/run-tests.sh
```diff
@@ -0,0 +1,49 @@
...
+[ -z "$MARCH" ] && MARCH=$(uname -m | tr '[:upper:]' '[:lower:]')
+
+if [ "$AUTH" != "noauth" ]; then
```
daprahamian: Can we take strings like this and stick them in variables? I'm worried about things like `noauth` or `nossl` being changed in the future.

.evg.yml
```diff
@@ -0,0 +1,266 @@
...
+          - windows-64-vs2013-compile
+          - windows-64-vs2013-test
+          - windows-64-vs2015-compile
+          - windows-64-vs2015-test
+          - windows-64-vs2015-small
+          - windows-64-vs2015-large
+  - id: node-version
+    display_name: "Node.js"
+    values:
+      - id: "Argon"
```
daprahamian: Are we not making one of these for `Carbon`?

.evg.yml
```diff
@@ -0,0 +1,266 @@
...
+        display_name: "Node.js Argon LTS"
+        variables:
+           NODE_LTS_NAME: "argon"
+      - id: "Boron"
+        display_name: "Node.js Boron LTS"
+        variables:
+           NODE_LTS_NAME: "boron"
+
+buildvariants:
```
daprahamian: Does evergreen not support >2D matrices? It would be nice to be able to add an axis for Mongo Version

.evergreen/config.yml
```diff
@@ -0,0 +1,629 @@
+# When a task that used to pass starts to fail
```
daprahamian: what is the difference between this file and `.evg.yml`?

lib/connection/pool.js
mbroadst: I was actually using this to determine when the pool was destroyed, before I opted to use connection spies. It's a harmless addition, but adds flexibility to the design (it's a no-op if nobody is listening) so I'd opt for keeping it in.

.evergreen/run-tests.sh
mbroadst: this is a straight rip of the existing mongo-orchestration scripts, soon to be replaced by work Sam Kleinman is doing with the `curator` method (read: this is going away soon).

test/config.js
mbroadst: I was actually glad to have found this, we often have cases where we `expect(err).to.not.exist`, however if it _does_ it shows something like `Expected { Object(namespace, message) } to be null` which is not helpful _at all_. The truncation configuration option here expands to show the full message. The other options here are just the defaults, but I'm otherwise codifying that we are indeed using the defaults.

test/tests/functional/basic_single_server_auth_tests.js
mbroadst: yeah I'm not sure what happened here, I think these were skipped from a while ago but sure I can `describe.skip` it.

.evergreen/run-tests.sh
mbroadst: I think it's a bit premature, there's only a single reference to it. I'll remove the whole section of code actually, since we aren't presently running auth or ssl tests and we aren't likely to use this pattern anyway.

.evergreen/config.yml
mbroadst: ack `.evg.yml` was accidentally commited. Removing now, and disregarding review from that file.

daprahamian: ok, then transplant all my comments from `.evg.yml` to `evergreen/config.yml`

.evg.yml
mbroadst: unfortunately not :'(

.evg.yml
mbroadst: Yes, Carbon is in the `.evergreen/config.yml`

mbroadst:
@daprahamian okay comments should be addressed

lib/connection/pool.js
daprahamian: Lets make sure that when we squash, this is a separate commit so we can track it.

.evergreen/config.yml
mbroadst: honestly I already squashed a whole bunch of commits here, I'm inclined to leave it as is

PR #289, opened 2018-04-02
NODE-1390 improve command monitoring implementation (CORE)

mbroadst:
This is the buddy implementation to https://github.com/mongodb/node-mongodb-native/pull/1675, see that PR for more details.

### DIFF
--- a/lib/connection/apm.js
+++ b/lib/connection/apm.js
@@ -23,6 +23,9 @@ const calculateDurationInMs = started => {
   return ((hrtime[0] * 1e9) + hrtime[1]) / 1e6;
 }
 
+const namespace = command => command.ns;
+const databaseName = command => command.ns.split('.')[0];
+const collectionName = command => command.ns.split('.')[1];
 const generateConnectionId = pool => `${pool.options.host}:${pool.options.port}`;
 const maybeRedact = (commandName, result) =>
   SENSITIVE_COMMANDS.has(commandName) ? {} : result;
@@ -66,21 +69,21 @@ const extractCommand = command => {
   if (command instanceof GetMore) {
     return {
       getMore: command.cursorId,
-      collection: command.ns.split('.')[1],
+      collection: collectionName(command),
       batchSize: command.numberToReturn
     };
   }
 
   if (command instanceof KillCursor) {
     return {
-      killCursors: command.ns.split('.')[1],
+      killCursors: collectionName(command),
       cursors: command.cursorIds
     };
   }
 
-  if (command.query && typeof command.query.$query !== 'undefined') {
+  if (command.query && command.query.$query) {
     // upconvert legacy find command
-    const result = { find: command.ns.split('.')[1] };
+    const result = { find: collectionName(command) };
     Object.keys(LEGACY_FIND_QUERY_MAP).forEach(key => {
       if (typeof command.query[key] !== 'undefined')
         result[LEGACY_FIND_QUERY_MAP[key]] = command.query[key];
@@ -115,7 +118,7 @@ const extractReply = (command, reply) => {
       ok: 1,
       cursor: {
         id: reply.message.cursorId,
-        ns: command.ns,
+        ns: namespace(command),
         nextBatch: reply.message.documents
       }
     };
@@ -134,7 +137,7 @@ const extractReply = (command, reply) => {
       ok: 1,
       cursor: {
         id: reply.message.cursorId,
-        ns: command.ns,
+        ns: namespace(command),
         firstBatch: reply.message.documents
       }
     };
@@ -163,7 +166,7 @@ class CommandStartedEvent {
 
     Object.assign(this, {
       command: cmd,
-      databaseName: command.ns.split('.')[0],
+      databaseName: databaseName(command),
       commandName,
       requestId: command.requestId,
       connectionId: generateConnectionId(pool)
--- a/lib/connection/apm.js
+++ b/lib/connection/apm.js
@@ -1,0 +1,105 @@
+'use strict';
+
+/** Commands that we want to redact because of the sensitive nature of their contents */
+const SENSITIVE_COMMANDS = [
+  'authenticate',
+  'saslStart',
+  'saslContinue',
+  'getnonce',
+  'createUser',
+  'updateUser',
+  'copydbgetnonce',
+  'copydbsaslstart',
+  'copydb'
+];
+
+// helper methods
+function extractCommand(command) { return command.query ? command.query : command; }
+function extractCommandName(command) { return Object.keys(command)[0]; }
+function calculateDuration(started) { return Date.now() - started; }
+function generateConnectionId(pool) { return `${pool.options.host}:${pool.options.port}`; }
+function maybeRedact(commandName, result) {
+  return SENSITIVE_COMMANDS.indexOf(commandName) !== -1 ? {} : result;
+}
+
+/** An event indicating the start of a given command */
+class CommandStartedEvent {
+  /**
+   * Create a started event
+   *
+   * @param {Pool} pool the pool that originated the command
+   * @param {Object} command the command
+   */
+  constructor (pool, command) {
+    const cmd = extractCommand(command);
+    const commandName = extractCommandName(cmd);
+
+    // NOTE: remove in major revision, this is not spec behavior
+    if (SENSITIVE_COMMANDS.indexOf(commandName) !== -1) {
+      this.commandObj = {};
+      this.commandObj[commandName] = true;
+    }
+
+    Object.assign(this, {
+      command: cmd,
+      databaseName: command.ns.split('.')[0],
+      commandName: extractCommandName(cmd),
+      requestId: command.requestId,
+      connectionId: generateConnectionId(pool)
+    });
+  }
+};
+
+/** An event indicating the success of a given command */
+class CommandSucceededEvent {
+  /**
+   * Create a succeeded event
+   *
+   * @param {Pool} pool the pool that originated the command
+   * @param {Object} command the command
+   * @param {Object} reply the reply for this command from the server
+   * @param {Number} started a timestamp of when the command was first sent to calculate duration
+   */
+  constructor(pool, command, reply, started) {
+    const cmd = extractCommand(command);
+    const commandName = extractCommandName(cmd);
+
+    Object.assign(this, {
+      duration: calculateDuration(started),
+      commandName,
+      reply: maybeRedact(commandName, reply.result),
+      requestId: command.requestId,
+      connectionId: generateConnectionId(pool)
+    });
+  }
+};
+
+/** An event indicating the failure of a given command */
+class CommandFailedEvent {
+  /**
+   * Create a failure event
+   *
+   * @param {Pool} pool the pool that originated the command
+   * @param {Object} command the command
+   * @param {MongoError|Object} error the generated error or a server error response
+   * @param {Number} started a timestamp of when the command was first sent to calculate duration
+   */
+  constructor(pool, command, error, started) {
+    const cmd = extractCommand(command);
+    const commandName = extractCommandName(cmd);
+
+    Object.assign(this, {
+      duration: calculateDuration(started),
+      commandName,
+      failure: maybeRedact(commandName, error),
+      requestId: command.requestId,
+      connectionId: generateConnectionId(pool)
+    });
+  }
+};
+
+module.exports = {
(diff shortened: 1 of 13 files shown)


lib/connection/apm.js
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  if (command instanceof KillCursor) {
+    return {
+      killCursors: command.ns.split('.')[1],
+      cursors: command.cursorIds
+    };
+  }
+
+  if (command.query && typeof command.query.$query !== 'undefined') {
```
daprahamian: Is the `typeof` necessary?

Can we do `if (command.query && command.query.$query !== undefined) {`?

lib/connection/apm.js
```diff
@@ -0,0 +1,221 @@
...
+    // upconvert legacy find command
+    const result = { find: command.ns.split('.')[1] };
+    Object.keys(LEGACY_FIND_QUERY_MAP).forEach(key => {
+      if (typeof command.query[key] !== 'undefined')
```
daprahamian: Same as above. Additionally, do we want `null` to be copied over?

lib/connection/apm.js
```diff
@@ -0,0 +1,221 @@
...
+  if (command instanceof KillCursor) {
+    return {
+      killCursors: command.ns.split('.')[1],
+      cursors: command.cursorIds
+    };
+  }
+
+  if (command.query && typeof command.query.$query !== 'undefined') {
+    // upconvert legacy find command
+    const result = { find: command.ns.split('.')[1] };
```
daprahamian: Since we appear to reuse it a lot, can we make a

```js
const getCollectionName = command => command.ns.split('.')[1];
```

lib/connection/apm.js
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    return result;
+  }
+
+  return command.query ? command.query : command;
+};
+
+const extractReply = (command, reply) => {
+  if (command instanceof GetMore) {
```
daprahamian: How many functions do we have of the form:

```js
const myFunction = (command, ...) => {
  if (command instanceof GetMore) {
    return ...;
  }
  if (command instanceof KillCursor) {
    return ...;
  }
  if (command.query && typeof command.query.$query !== 'undefined') {
    return ...;
  }
  return ...;
};
```

Is this a pattern that we'd want to codify, or would that be over-optimizing?

Something like:

```js
const GET_MORE_COMMAND = Symbol();
const KILL_CURSOR_COMMAND = Symbol();
const FIND_COMMAND = Symbol();
const OTHER_COMMAND = Symbol();
const typeOfCommand = command => {
  if (command instanceof GetMore) {
    return GET_MORE_COMMAND;
  }
  if (command instanceof KillCursor) {
    return KILL_CURSOR_COMMAND;
  }
  if (command.query && typeof command.query.$query !== 'undefined') {
    return FIND_COMMAND;
  }
  return OTHER_COMMAND;
};

const extractCommand = command  => {
  switch (typeOfCommand(command)) {
    case GET_MORE_COMMAND: return ...;
    case KILL_CURSOR_COMMAND: return ...;
    case FIND_COMMAND: return ...;
    default: return ...;
  }
}
```

lib/connection/apm.js
```diff
@@ -0,0 +1,221 @@
...
+        firstBatch: reply.message.documents
+      }
+    };
+  }
+
+  return reply.result;
+};
+
+/** An event indicating the start of a given command */
+class CommandStartedEvent {
```
daprahamian: `CommandStartedEvent`, `CommandSucceededEvent` and `CommandFailedEvent` seem to have a lot in common. Can we make a common base class and/or use composition to reduce the duplicate logic?

lib/connection/apm.js
```diff
@@ -0,0 +1,221 @@
...
+class CommandSucceededEvent {
+  /**
+   * Create a succeeded event
+   *
+   * @param {Pool} pool the pool that originated the command
+   * @param {Object} command the command
+   * @param {Object} reply the reply for this command from the server
+   * @param {Number} started a timestamp of when the command was first sent to calculate duration
+   */
+  constructor(pool, command, reply, started) {
```
daprahamian: Can we pass an `options` object instead of 4 params?

lib/connection/apm.js
```diff
@@ -0,0 +1,221 @@
...
+class CommandFailedEvent {
+  /**
+   * Create a failure event
+   *
+   * @param {Pool} pool the pool that originated the command
+   * @param {Object} command the command
+   * @param {MongoError|Object} error the generated error or a server error response
+   * @param {Number} started a timestamp of when the command was first sent to calculate duration
+   */
+  constructor(pool, command, error, started) {
```
daprahamian: Same as above

lib/connection/apm.js
daprahamian: Additionally, do these have to be classes? All they have is constructors. It seems like we could instead have factory functions that return POJOs.

lib/connection/commands.js
```diff
@@ -53,6 +53,9 @@ var Query = function(bson, ns, query, options) {
   this.returnFieldSelector = options.returnFieldSelector || null;
   this.requestId = Query.getRequestId();
 
+  // special case for pre-3.2 find commands, delete ASAP
+  this.pre32Limit = options.pre32Limit || null;
```
daprahamian: Is there a reason we are using `null` here, and not `undefined`?

lib/connection/pool.js
```diff
@@ -1226,6 +1228,38 @@ Pool.prototype.write = function(commands, options, cb) {
     });
   }
 
+  // If command monitoring is enabled we need to modify the callback here
+  if (self.options.enableCommandMonitoring) {
+    // NOTE: there is only ever a single command, for some legay reason I am unaware of we
```
daprahamian: `s/legay/legacy/`

lib/topologies/mongos.js
```diff
@@ -467,6 +468,12 @@ function connectProxies(self, servers) {
       server.once('parseError', handleInitialConnectEvent(self, 'parseError'));
       server.once('error', handleInitialConnectEvent(self, 'error'));
       server.once('connect', handleInitialConnectEvent(self, 'connect'));
+
+      // Command Monitoring events
+      server.on('commandStarted', event => self.emit('commandStarted', event));
```
daprahamian: Can we make a shared function like this:

```js
function addCommandMonitoringListeners(listener, emitter) {
  ['commandStarted', 'commandSucceeded', 'commandFailed'].forEach(eventName => {
    listener.on(eventName, event => emitter.emit(eventName, event));
  });
}
```

lib/wireprotocol/2_6_support.js
```diff
@@ -104,12 +104,13 @@ WireProtocol.prototype.killCursor = function(bson, ns, cursorState, pool, callba
     try {
       pool.write(killCursor, options, callback);
     } catch (err) {
-      callback(err, null);
+      if (typeof callback === 'function') {
+        callback(err, null);
+      } else {
+        console.warn(err);
```
daprahamian: should we be `console.warn`ing here? I assume any low-level wire protocol errors, should be logged to a `Logger` object.

lib/wireprotocol/2_6_support.js
```diff
@@ -216,41 +217,33 @@ var setupClassicFind = function(bson, ns, cmd, cursorState, topology, options) {
...
-  if (cmd.hint) (findCmd['$hint'] = cmd.hint), (usesSpecialModifier = true);
-  if (cmd.snapshot) (findCmd['$snapshot'] = cmd.snapshot), (usesSpecialModifier = true);
-  if (cmd.returnKey) (findCmd['$returnKey'] = cmd.returnKey), (usesSpecialModifier = true);
-  if (cmd.maxScan) (findCmd['$maxScan'] = cmd.maxScan), (usesSpecialModifier = true);
-  if (cmd.min) (findCmd['$min'] = cmd.min), (usesSpecialModifier = true);
-  if (cmd.max) (findCmd['$max'] = cmd.max), (usesSpecialModifier = true);
-  if (cmd.showDiskLoc) (findCmd['$showDiskLoc'] = cmd.showDiskLoc), (usesSpecialModifier = true);
-  if (cmd.comment) (findCmd['$comment'] = cmd.comment), (usesSpecialModifier = true);
-  if (cmd.maxTimeMS) (findCmd['$maxTimeMS'] = cmd.maxTimeMS), (usesSpecialModifier = true);
+  if (cmd.sort) findCmd['$orderby'] = cmd.sort;
```
daprahamian: Nice. :+1:

lib/connection/apm.js
```diff
@@ -0,0 +1,221 @@
+'use strict';
+const KillCursor = require('../connection/commands').KillCursor;
+const GetMore = require('../connection/commands').GetMore;
+
+/** Commands that we want to redact because of the sensitive nature of their contents */
+const SENSITIVE_COMMANDS = [
```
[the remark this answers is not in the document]
mbroadst: Okay, I made it a `Set`, but I think `isSensitiveCommand` is getting a bit pedantic 😄

lib/connection/apm.js
```diff
@@ -0,0 +1,221 @@
...
+  'createUser',
+  'updateUser',
+  'copydbgetnonce',
+  'copydbsaslstart',
+  'copydb'
+];
+
+// helper methods
+const extractCommandName = command => Object.keys(command)[0];
+const calculateDuration = started => Date.now() - started;
```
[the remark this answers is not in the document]
mbroadst: `process.hrtime` is fine by me, I was trying to keep parity with existing APM code but the duration is not likely to be something people are checking for strictness

lib/connection/apm.js
mbroadst: in this case yes I think we can just do `if (command.query && command.query.$query)`

lib/connection/apm.js
mbroadst: in this case, I prefer this form so we can potentially catch booleans.

lib/connection/apm.js
mbroadst: sure

lib/connection/apm.js
mbroadst: we only have them here, and only as long as we support the old op codes (which are in the process of dying). I don't think there's compelling reason to further refactor this at this time.

lib/connection/apm.js
mbroadst: The original implementation returned POJOs, but I preferred them being well named/defined types. Do you have a case for them being POJOs?

As for duplication, I suppose we could make a common `CommandMonitoringEvent`, but I'm not sure we gain all that much here. I like that I can look at each of these definitions and very clearly see what they doing. I am generally on your side of the fence on issues like these, but when it comes to types like events or errors I prefer serious verbosity (for readabilities sake)

lib/connection/apm.js
mbroadst: I'm not sure there's a compelling reason for this, as it is private API, is there?

lib/connection/commands.js
mbroadst: I think simply for the sake of defining it with a value. I always feel weird assigning undefined to a property, as the intent would be to indicate it doesn't exist. Perhaps a better idea here is to do:
```
if (options.pre32Limit) this.pre32Limit = options.pre32Limit;
```

lib/topologies/mongos.js
mbroadst: sure

lib/wireprotocol/2_6_support.js
mbroadst: Generally: yes. Here's my argument for this one: this is explicitly the case where you call killCursors (a legacy command people are not likely using anymore) without a callback (which we expressly recommend against). Point being, I'd like to fix this if and when someone actually complains about it, because we don't easily have access to a defined logger here and it's a very niche situation.

lib/connection/commands.js
daprahamian: Ok. In general, I'd like to not use `null` as it causes a lot of problems in JS. We could just do:

```js
this.pre32Limit = options.pre32Limit;
```

Also, if someone passes in `{ pre32Limit: 0 }`, the current behavior will set `this.pre32Limit` to `null`, not `0`. Do we want that behavior?

mbroadst:
```
this.pre32Limit = options.pre32Limit;
```

sounds good to me. setting it to null is fine.

PR #294, opened 2018-04-12
fix(auth): prevent stalling on authentication when connected

daprahamian:
ReplicaSet was storing the authentication call in the disconnectHandler
when it didn't need to, leading to auth stalling. From now on, will
only store handlers if ReplSet is not actually connected.

@durran Can you link against this and verify that it fixes your issue?


mbroadst:
@durran rad, @daprahamian great job!

PR #295, opened 2018-04-17
feat(transactions): add initial support for multi-stage transactions on replicasets

mbroadst:
These are the changes required for the core driver in order to support transactions in the mongodb node driver. The meat of the changes exist in the `ClientSession` object, as well as a smattering of logical changes related to whether a given session is currently in a transaction. Please try to focus on functional criticisms, I guarantee you if you're having thoughts about overall architecture I have probably had nightmares about them in the past few weeks

### DIFF
--- a/lib/sessions.js
+++ b/lib/sessions.js
@@ -34,6 +34,7 @@ class ClientSession extends EventEmitter {
    * @param {Boolean} [options.causalConsistency] Whether causal consistency should be enabled on this session
    * @param {Boolean} [options.autoStartTransaction=false] When enabled this session automatically starts a transaction with the provided defaultTransactionOptions.
    * @param {Object} [options.defaultTransactionOptions] The default TransactionOptions to use for transactions started on this session.
+   * @param {Object} [clientOptions] Optional settings provided when creating a client in the porcelain driver
    */
   constructor(topology, sessionPool, options, clientOptions) {
     super();
@@ -134,6 +135,23 @@ class ClientSession extends EventEmitter {
     }
 
     return this.id.id.buffer.equals(session.id.id.buffer);
+  }
+
+  /**
+   * Increment the transaction number on the internal ServerSession
+   */
+  incrementTransactionNumber() {
+    this.serverSession.txnNumber++;
+  }
+
+  /**
+   * Increment the statement id on the internal ServerSession
+   *
+   * @param {Number} [operationCount] the number of operations performed
+   */
+  incrementStatementId(operationCount) {
+    operationCount = operationCount || 1;
+    this.serverSession.stmtId += operationCount;
   }
 
   /**
@@ -265,7 +283,7 @@ function endTransaction(clientSession, commandName, callback) {
   const command = { [commandName]: 1 };
   if (clientSession.transactionOptions.writeConcern) {
     Object.assign(command, { writeConcern: clientSession.transactionOptions.writeConcern });
-  } else if (clientSession.clientOptions.w) {
+  } else if (clientSession.clientOptions && clientSession.clientOptions.w) {
     Object.assign(command, { writeConcern: { w: clientSession.clientOptions.w } });
   }
 
--- a/lib/sessions.js
+++ b/lib/sessions.js
@@ -244,7 +244,7 @@ function endTransaction(clientSession, commandName, callback) {
   if (clientSession.transactionOptions.writeConcern) {
     Object.assign(command, { writeConcern: clientSession.transactionOptions.writeConcern });
   } else if (clientSession.clientOptions.w) {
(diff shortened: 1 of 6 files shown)


lib/wireprotocol/3_2_support.js
```diff
@@ -1,19 +1,67 @@
...
+ */
+function decorateWithTransactionsData(command, session) {
+  if (!session) {
+    return;
+  }
+
+  // first apply non-transaction-specific sessions data
+  const serverSession = session.serverSession;
+  if (serverSession.txnNumber) {
+    command.txnNumber = BSON.Long.fromNumber(serverSession.txnNumber);
```
daprahamian: do we need to use a specific BSON impl?

lib/sessions.js
```diff
@@ -1,16 +1,41 @@
...
+   * WARNING: not meant to be instantiated directly
+   *
+   * @param {Topology} topology The current client's topology
+   * @param {ServerSessionPool} sessionPool The server session pool
+   * @param {Object} [options] Optional settings
+   * @param {Boolean} [options.causalConsistency] Whether causal consistency should be enabled on this session
+   * @param {Boolean} [options.autoStartTransaction=false] When enabled this session automatically starts a transaction with the provided defaultTransactionOptions.
+   * @param {Object} [options.defaultTransactionOptions] The default TransactionOptions to use for transactions started on this session.
+   */
+  constructor(topology, sessionPool, options, clientOptions) {
```
daprahamian: `clientOptions` needs some jsdoc

lib/sessions.js
```diff
@@ -98,6 +135,167 @@ class ClientSession extends EventEmitter {
...
+    // The server transaction was never started.
+    resetTransactionState(clientSession);
+    callback(null, null);
+    return;
+  }
+
+  const command = { [commandName]: 1 };
+  if (clientSession.transactionOptions.writeConcern) {
+    Object.assign(command, { writeConcern: clientSession.transactionOptions.writeConcern });
+  } else if (clientSession.clientOptions.w) {
```
daprahamian: as far as I can tell, we don't have a guarantee that `clientSession.clientOptions != null`, so we might want to guard this.

lib/topologies/shared.js
```diff
@@ -425,10 +422,20 @@ const isRetryableWritesSupported = function(topology) {
  *
  * @param {ClientSession} session
  */
-const getNextTransactionNumber = function(session) {
+function incrementTransactionNumber(session) {
```
daprahamian: is there a reason this cannot be a member function on a session?

lib/topologies/shared.js
```diff
@@ -425,10 +422,20 @@ const isRetryableWritesSupported = function(topology) {
...
-};
+}
+
+/**
+ * Increment the statement id on the ServerSession contained by the provided ClientSession
+ *
+ * @param {ClientSession} session the client sessions
+ * @param {Number} [operationCount] the number of operations performed
+ */
+function incrementStatementId(session, operationCount) {
```
daprahamian: same for this guy

lib/wireprotocol/3_2_support.js
```diff
@@ -1,19 +1,67 @@
...
+  }
+
+  command.stmtId = serverSession.stmtId;
+  command.autocommit = false;
+
+  if (serverSession.stmtId === 0) {
+    command.startTransaction = true;
+    command.readConcern = { level: 'snapshot' };
+
+    if (session.transactionOptions.readConcern) {
```
daprahamian:
```js
const readConcern = session.transactionOptions.readConcern || session.clientOptions.readConcern;
if (readConcern) {
  command.readConcern = readConcern;
}
```

Also may need to guard against `transactionOptions` and/or `clientOptions` not being there.

lib/wireprotocol/3_2_support.js
```diff
@@ -1,19 +1,67 @@
...
+      Object.assign(command, { readConcern: session.transactionOptions.readConcern });
+    } else if (session.clientOptions.readConcern) {
+      Object.assign(command, { readConcern: session.clientOptions.readConcern });
+    }
+
+    if (session.supports.causalConsistency && session.operationTime) {
+      Object.assign(command.readConcern, { afterClusterTime: session.operationTime });
```
daprahamian:
```js
command.readConcern.afterClusterTime = session.operationTime
```

Also assumes that `command.readConcern` is a thing.

lib/wireprotocol/3_2_support.js
mbroadst: I do find this very confusing about the codebase in general, but I believe no. In many places we pass in a `bson` object, but also the `BSON` we are using here is the result of `retrieveBSON` above - so we are allowing users to provide their own BSON library, just not by passing it in to the topologies.

PR #296, opened 2018-04-18
docs(sessions): add missing default for `causalConsistency`

vkarpov15:
I'm pretty sure that `causalConsistency` is false by default but I might be wrong. Also, see https://github.com/mongodb/node-mongodb-native/pull/1682


mbroadst:
@vkarpov15 ah, actually causal is `true` by default (see: [here](https://github.com/mongodb-js/mongodb-core/blob/master/lib/sessions.js#L59))

vkarpov15:
My mistake, thanks for pointing that out. I'll close this :+1:

PR #298, opened 2018-04-19
fix(pool): ensure that lsid is sent in get requests to mongos

daprahamian:
There is a strange case, when topology is mongos and readPreference
is not primary, where we nest our find queries in a "$query" attribute.
This was having the effect of not sending the lsid on find requests,
which caused further errors when getMores attempted to send an lsid.

Fixes NODE-1420


mbroadst:
@daprahamian please remember to cherry-pick this to 3.0, we should push a 3.0.8 release soon with this as well as other fixes (not constantly sending `endSessions` for instance)

PR #300, opened 2018-04-23
Adds saslprep and SCRAM-SHA-256

daprahamian:
No longer WIP.

Includes SCRAM-SHA-256, saslprep for Node >= 6, and Mechanism Negotiation.

Fixes NODE-1303
Fixes NODE-1311

### DIFF
--- a/lib/auth/scram.js
+++ b/lib/auth/scram.js
@@ -155,12 +155,11 @@ ScramSHA.prototype.auth = function(server, connections, db, username, password, 
   var numberOfValidConnections = 0;
   var errorObject = null;
 
-  let cryptoMethod = 'sha1';
+  let cryptoMethod = this.cryptoMethod;
   let mechanism = 'SCRAM-SHA-1';
   let processedPassword;
 
   if (this.cryptoMethod === 'sha256') {
-    cryptoMethod = 'sha256';
     mechanism = 'SCRAM-SHA-256';
 
     let saslprepFn = (server.s && server.s.saslprep) || saslprep;
@@ -187,17 +186,17 @@ ScramSHA.prototype.auth = function(server, connections, db, username, password, 
     // NOTE: This is done b/c Javascript uses UTF-16, but the server is hashing in UTF-8.
     // Since the username is not sasl-prep-d, we need to do this here.
     const firstBare = Buffer.concat([
-      Buffer.from('n='),
-      Buffer.from(username),
-      Buffer.from(',r='),
-      Buffer.from(nonce)
+      Buffer.from('n=', 'utf8'),
+      Buffer.from(username, 'utf8'),
+      Buffer.from(',r=', 'utf8'),
+      Buffer.from(nonce, 'utf8')
     ]);
 
     // Build command structure
     var cmd = {
       saslStart: 1,
       mechanism: mechanism,
-      payload: new Binary(Buffer.concat([Buffer.from('n,,'), firstBare])),
+      payload: new Binary(Buffer.concat([Buffer.from('n,,', 'utf8'), firstBare])),
       autoAuthorize: 1
     };
 
--- a/lib/auth/scram.js
+++ b/lib/auth/scram.js
@@ -6,6 +6,14 @@ var f = require('util').format,
   Query = require('../connection/commands').Query,
   MongoError = require('../error').MongoError;
 
+let saslprep;
+
+try {
+  saslprep = require('saslprep');
+} catch (e) {
+  // don't do anything;
+}
+
 var BSON = retrieveBSON(),
   Binary = BSON.Binary;
 
@@ -26,14 +34,15 @@ AuthSession.prototype.equal = function(session) {
 var id = 0;
 
 /**
- * Creates a new ScramSHA1 authentication mechanism
+ * Creates a new ScramSHA authentication mechanism
  * @class
- * @return {ScramSHA1} A cursor instance
+ * @return {ScramSHA} A cursor instance
  */
-var ScramSHA1 = function(bson) {
+var ScramSHA = function(bson, cryptoMethod) {
   this.bson = bson;
   this.authStore = [];
   this.id = id++;
+  this.cryptoMethod = cryptoMethod || 'sha1';
 };
 
 var parsePayload = function(payload) {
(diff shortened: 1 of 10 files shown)


lib/auth/scram.js
```diff
@@ -124,6 +155,26 @@ ScramSHA1.prototype.auth = function(server, connections, db, username, password,
   var numberOfValidConnections = 0;
   var errorObject = null;
 
+  let cryptoMethod = 'sha1';
+  let mechanism = 'SCRAM-SHA-1';
+  let processedPassword;
+
+  if (this.cryptoMethod === 'sha256') {
+    cryptoMethod = 'sha256';
```
mbroadst: can't you just use `this.cryptoMethod` later rather than creating a new variable?

lib/auth/scram.js
```diff
@@ -132,13 +183,21 @@ ScramSHA1.prototype.auth = function(server, connections, db, username, password,
     // Create a random nonce
     var nonce = crypto.randomBytes(24).toString('base64');
     // var nonce = 'MsQUY9iw0T9fx2MUEz6LZPwGuhVvWAhc'
-    var firstBare = f('n=%s,r=%s', username, nonce);
+
+    // NOTE: This is done b/c Javascript uses UTF-16, but the server is hashing in UTF-8.
+    // Since the username is not sasl-prep-d, we need to do this here.
+    const firstBare = Buffer.concat([
```
mbroadst: do we want to further explicitly state the encoding as `utf-8` even though that's the default?

lib/topologies/server.js
```diff
@@ -21,6 +21,57 @@ var inherits = require('util').inherits,
   SessionMixins = require('./shared').SessionMixins,
   relayEvents = require('./shared').relayEvents;
 
+function getSaslSupportedMechs(options) {
+  if (!options) {
+    return {};
+  }
+
+  const authArray = options.auth || [];
+  const authMechanism = authArray[0] || options.authMechanism;
```
mbroadst: This looks like it can still potentially be `undefined`

lib/topologies/server.js
```diff
@@ -21,6 +21,57 @@ var inherits = require('util').inherits,
...
+    if (ismaster.maxWireVersion >= 3) {
+      return 'scram-sha-1';
+    }
+  }
+
+  // Default for wireprotocol < 3
+  return 'mongocr';
+}
+
+function parseOutIsMasterError(err, result) {
```
mbroadst: nit: is it really parsing, or extracting?

test/tests/functional/basic_single_server_auth_tests.js
```diff
@@ -329,7 +381,7 @@ describe.skip('Basic single server auth tests', function() {
     }
   });
 
-  it('should correctly authenticate server using scram-sha-1 using connect auth then logout', {
+  it.skip('should correctly authenticate server using scram-sha-1 using connect auth then logout', {
```
mbroadst: what's going on here? should we perhaps skip it only for <3.7.3?

test/tests/functional/basic_single_server_auth_tests.js
```diff
@@ -8,7 +8,59 @@ var expect = require('chai').expect,
...
+          bson: new Bson()
+        });
+
+        server.on('connect', _server => {
+          dropUser((dropUserErr, dropUserRes) => {
+            expect(dropUserErr).to.be.null;
+            expect(dropUserRes).to.exist;
+
+            _server.destroy({ force: true });
+            expect(Object.keys(Connection.connections()).length).to.equal(0);
```
mbroadst: I've been trying to remove all traces of these types of tests. This was here to ensure that all connections were really closed before moving on to the next test, and we have better ways of doing this now with the `ConnectionSpy` (at least). Ideally, `destroy` would take a callback but I just made [a ticket](https://jira.mongodb.org/browse/NODE-1443) for that

lib/topologies/server.js
daprahamian: It is possible for it to be `undefined`. That's one of the cases that we want to handle with the `saslSupportedMechs`

lib/auth/scram.js
daprahamian: I had something in mind about decoupling the symbol we use from the nodejs value. If we wanted to switch to `Symbol`s for example. But you're right, it isn't necessary.

lib/auth/scram.js
daprahamian: Yeah, I can add that in.

test/tests/functional/basic_single_server_auth_tests.js
daprahamian: This  test is just broken in general, before I even made changes.

test/tests/functional/basic_single_server_auth_tests.js
daprahamian: Ok, I'll remove it.

lib/auth/scram.js
```diff
@@ -124,6 +155,25 @@ ScramSHA1.prototype.auth = function(server, connections, db, username, password,
   var numberOfValidConnections = 0;
   var errorObject = null;
 
+  let cryptoMethod = this.cryptoMethod;
```
mbroadst: wow I thought this was unused, but this method is like hundreds of lines long (!)

lib/auth/scram.js
```diff
@@ -124,6 +155,25 @@ ScramSHA1.prototype.auth = function(server, connections, db, username, password,
...
+  let mechanism = 'SCRAM-SHA-1';
+  let processedPassword;
+
+  if (this.cryptoMethod === 'sha256') {
```
mbroadst: nit: you can definitely reuse `cryptoMethod` here though

lib/topologies/mongos.js
```diff
@@ -37,12 +37,7 @@ const BSON = retrieveBSON();
  * server.connect();
  */
 
-var MongoCR = require('../auth/mongocr'),
-  X509 = require('../auth/x509'),
-  Plain = require('../auth/plain'),
-  GSSAPI = require('../auth/gssapi'),
-  SSPI = require('../auth/sspi'),
-  ScramSHA1 = require('../auth/scram');
+const defaultAuthProviders = require('../auth/defaultAuthProviders').defaultAuthProviders;
```
mbroadst: love these cleanups!

package.json
```diff
@@ -44,5 +44,8 @@
   "bugs": {
     "url": "https://github.com/mongodb-js/mongodb-core/issues"
   },
-  "homepage": "https://github.com/mongodb-js/mongodb-core"
+  "homepage": "https://github.com/mongodb-js/mongodb-core",
+  "optionalDependencies": {
```
mbroadst: should we switch all of our `peerOptionalDependencies` to just be `optionalDependencies` 🤔

mbroadst: I just went and saw a number of issues users were having with `require_optional` (e.g. [this](https://github.com/christkv/require_optional/issues/3)), maybe we should discuss removing it in favor of just `optionalDependencies`

daprahamian:
PR Updated with feedback

mbroadst:
:shipit:

PR #303, opened 2018-04-30
NODE-1259 Refactor mongodb-core to use a single Topology type (part 1)

mbroadst:
This PR introduces a single `Topology` type meant to completely replace `Server`/`Replset`/`Mongos` in both `mongodb-core` and `node-mongodb-native`. Presently it is not hooked up at all to the driver, but passes all SDAM, server selection and max-staleness tests. The next step will be to actually connect the internal pieces so we can effectively replace the existing topologies. (Broken apart to give a glimmer of a chance at reviewing)


lib/sdam/topology_description.js
```diff
@@ -0,0 +1,352 @@
...
+const ServerDescription = require('./server_description').ServerDescription;
+const ReadPreference = require('../topologies/read_preference');
+
+// contstants related to compatability checks
+const MIN_SUPPORTED_SERVER_VERSION = '2.6';
+const MIN_SUPPORTED_WIRE_VERSION = 2;
+const MAX_SUPPORTED_WIRE_VERSION = 5;
+
+// An enumeration of topology types we know about
+const TopologyType = {
```
daprahamian:
```js
['Single', 'ReplicaSetNoPrimary', 'ReplicaSetWithPrimary', 'Sharded', 'Unknown'].reduce((memo, key) => {
  memo[key] = key;
  return memo;
}, {});
```

lib/sdam/topology_description.js
```diff
@@ -0,0 +1,352 @@
...
+  /**
+   * Create a TopologyDescription
+   *
+   * @param {string} topologyType
+   * @param {Map<string, ServerDescription>} serverDescriptions the a map of address to ServerDescription
+   * @param {string} setName
+   * @param {number} maxSetVersion
+   * @param {ObjectId} maxElectionId
+   */
+  constructor(topologyType, serverDescriptions, setName, maxSetVersion, maxElectionId, options) {
```
daprahamian: We should POJO this constructor.

lib/sdam/topology_description.js
```diff
@@ -0,0 +1,352 @@
...
+    this.localThresholdMS = options.localThresholdMS || 0;
+    this.options = options;
+
+    // determine server compatibility
+    for (const serverDescription of this.servers.values()) {
+      if (serverDescription.minWireVersion > MAX_SUPPORTED_WIRE_VERSION) {
+        this.compatible = false;
+        this.compatibilityError = `Server at ${serverDescription.address} requires wire version ${
+          serverDescription.minWireVersion
+        }, but this version of the driver only supports up to ${MAX_SUPPORTED_WIRE_VERSION}.`;
```
daprahamian: should there be a `break` after this?

lib/sdam/topology_description.js
```diff
@@ -0,0 +1,352 @@
...
+      }
+    }
+
+    // Whenever a client updates the TopologyDescription from an ismaster response, it MUST set
+    // TopologyDescription.logicalSessionTimeoutMinutes to the smallest logicalSessionTimeoutMinutes
+    // value among ServerDescriptions of all data-bearing server types. If any have a null
+    // logicalSessionTimeoutMinutes, then TopologyDescription.logicalSessionTimeoutMinutes MUST be
+    // set to null.
+    const readableServers = Array.from(this.servers.values()).filter(s => s.isReadable);
+    this.logicalSessionTimeoutMinutes = readableServers.reduce((result, server) => {
```
daprahamian: If I am reading this correctly, we want a set of [{}, {logicalSessionTimeoutMinutes: 5}] to return `null`, but this logic will return `5`.

daprahamian: I think we could do

```js
const serverMinutes = Array.from(this.servers.values()).filter(s => s.isReadable).map(s => s.logicalSessionTimeoutMinutes);
this.logicalSessionTimeoutMinutes = serverMinutes.findIndex(s => s == null) ? null : Math.min.apply(null, serverMinutes);
```

lib/sdam/topology_description.js
```diff
@@ -0,0 +1,352 @@
...
+      if (result == null) return server.logicalSessionTimeoutMinutes;
+      return Math.min(result, server.logicalSessionTimeoutMinutes);
+    }, null);
+  }
+
+  /**
+   * @returns The minimum reported wire version of all known servers
+   */
+  get commonWireVersion() {
+    return Array.from(this.servers.values())
```
daprahamian: Can we make a getter for `Array.from(this.servers.values())`?

daprahamian: Also:

```js
return Math.min.apply(null, 
  Array.from(this.servers.values())
    .filter(server => server.type !== ServerType.Unknown && server.maxWireVersion != null)
);
```

lib/sdam/topology_description.js
```diff
@@ -0,0 +1,352 @@
...
+        null
+      );
+  }
+
+  /**
+   * Returns a copy of this description updated with a given ServerDescription
+   *
+   * @param {ServerDescription} serverDescription
+   */
+  update(serverDescription) {
```
daprahamian: Not that I want to, but do we need to defensively code this to have a default value?

lib/sdam/topology_description.js
```diff
@@ -0,0 +1,352 @@
...
+      if (serverType === ServerType.RSPrimary) {
+        const result = updateRsFromPrimary(
+          serverDescriptions,
+          setName,
+          serverDescription,
+          maxSetVersion,
+          maxElectionId
+        );
+
+        (topologyType = result[0]),
```
daprahamian: Can these be separate statements?

lib/sdam/topology_description.js
```diff
@@ -0,0 +1,352 @@
...
+    let maxSetVersion = this.maxSetVersion;
+    let maxElectionId = this.maxElectionId;
+
+    const serverType = serverDescription.type;
+    let serverDescriptions = new Map(this.servers);
+
+    // update the actual server description
+    serverDescriptions.set(address, serverDescription);
+
+    if (topologyType === TopologyType.Single) {
```
daprahamian: For this whole function, can we make some sort of POJO map of `topologyType` to state transform, and invoke that? This is quite a long function

lib/sdam/topology_description.js
```diff
@@ -0,0 +1,352 @@
...
+   * following section for behaviour rules.
+   *
+   * @return {Boolean} Whether there is a writable server in this topology
+   */
+  hasWritableServer() {
+    return this.hasReadableServer(ReadPreference.primary);
+  }
+}
+
+function topologyTypeForServerType(serverType) {
```
daprahamian: Do we not do standalone in here?

lib/sdam/topology_description.js
```diff
@@ -0,0 +1,352 @@
...
+  if (serverType === ServerType.Mongos) return TopologyType.Sharded;
+  if (serverType === ServerType.RSPrimary) return TopologyType.ReplicaSetWithPrimary;
+  return TopologyType.ReplicaSetNoPrimary;
+}
+
+function updateRsFromPrimary(
```
daprahamian: It seems like this function should be returning a POJO with named fields instead of an array

lib/sdam/topology_description.js
```diff
@@ -0,0 +1,352 @@
...
+  const responseAddresses = serverDescription.allHosts;
+  currentAddresses.filter(addr => responseAddresses.indexOf(addr) === -1).forEach(address => {
+    serverDescriptions.delete(address);
+  });
+
+  return [checkHasPrimary(serverDescriptions), setName, maxSetVersion, maxElectionId];
+}
+
+function updateRsWithPrimaryFromMember(serverDescriptions, setName, serverDescription) {
+  assert.ok(setName, 'setName is required');
```
daprahamian: Shouldn't we do a `TypeError` here?

lib/sdam/topology_description.js
```diff
@@ -0,0 +1,352 @@
...
+
+  if (setName !== serverDescription.setName) {
```
daprahamian: Can it be a single `if` since the result is the same?

lib/sdam/topology_description.js
```diff
@@ -0,0 +1,352 @@
...
+    serverDescriptions.delete(serverDescription.address);
+  } else if (serverDescription.me && serverDescription.address !== serverDescription.me) {
+    serverDescriptions.delete(serverDescription.address);
+  }
+
+  return checkHasPrimary(serverDescriptions);
+}
+
+function updateRsNoPrimaryFromMember(serverDescriptions, setName, serverDescription) {
```
daprahamian: Return a POJO instead of an array

lib/sdam/server_description.js
```diff
@@ -0,0 +1,118 @@
...
+  constructor(address, ismaster, options) {
+    options = options || {};
+
+    this.address = address;
+    this.error = null;
+    this.roundTripTime = options.roundTripTime || 0;
+    this.lastWriteDate = ismaster && ismaster.lastWrite ? ismaster.lastWrite.lastWriteDate : null;
+    this.opTime = ismaster && ismaster.lastWrite ? ismaster.lastWrite.opTime : null;
+    this.type = parseServerType(ismaster);
+    this.minWireVersion = (ismaster && ismaster.minWireVersion) || 0;
```
daprahamian: I feel like a lot of these could be accomplished with an:

```js
Object.assign(this, { ...defaultValues }, ismaster);
```

lib/sdam/server_description.js
```diff
@@ -0,0 +1,118 @@
...
+  get isReadable() {
+    return this.type === ServerType.RSSecondary || this.isWritable;
+  }
+
+  /**
+   * @return {Boolean} Is this server available for writes
+   */
+  get isWritable() {
+    return (
+      [ServerType.RSPrimary, ServerType.Standalone, ServerType.Mongos].indexOf(this.type) !== -1
```
daprahamian: Can we do this with a `Set`?

lib/sdam/server_selectors.js
```diff
@@ -0,0 +1,214 @@
...
+const ServerType = require('./server_description').ServerType;
+const TopologyType = require('./topology_description').TopologyType;
+const ReadPreference = require('../topologies/read_preference');
+const MongoError = require('../error').MongoError;
+
+// max staleness constants
+const IDLE_WRITE_PERIOD = 10000;
+const SMALLEST_MAX_STALENESS_SECONDS = 90;
+
+function writableServerSelector() {
```
daprahamian: Is there a reason that we are returning a function here? There doesn't appear to be any closure at work here.

lib/sdam/server_selectors.js
```diff
@@ -0,0 +1,214 @@
...
+      const staleness = stalenessMS / 1000;
+      if (staleness <= readPreference.maxStalenessSeconds) result.push(server);
+      return result;
+    }, []);
+  }
+
+  return servers;
+}
+
+function tagSetMatch(tagSet, serverTags) {
```
daprahamian: Is there not a good `Set` method to do this?

lib/sdam/server_selectors.js
```diff
@@ -0,0 +1,214 @@
...
+    }
+  }
+
+  return [];
+}
+
+function latencyWindowReducer(topologyDescription, servers) {
+  const low = servers.reduce(
+    (min, server) =>
+      min === -1 ? server.roundTripTime : server.roundTripTime < min ? server.roundTripTime : min,
```
daprahamian: Nested ternaries hurt my brain. Can we unroll this?

lib/sdam/server_selectors.js
```diff
@@ -0,0 +1,214 @@
...
+
+    if (
+      topologyDescription.type === TopologyType.Single ||
+      topologyDescription.type === TopologyType.Sharded
+    ) {
+      return latencyWindowReducer(topologyDescription, servers.filter(knownFilter));
+    }
+
+    if (readPreference.mode === ReadPreference.PRIMARY) {
+      return servers.filter(s => s.type === ServerType.RSPrimary);
```
daprahamian: can we use `primaryFilter`?

lib/sdam/topology.js
```diff
@@ -0,0 +1,416 @@
...
+   * @param {Number} [options.serverSelectionTimeoutMS=30000] How long to block for server selection before throwing an error
+   * @param {Number} [options.heartbeatFrequencyMS=10000] The frequency with which topology updates are scheduled
+   */
+  constructor(seedlist, options) {
+    super();
+    seedlist = seedlist || [];
+    options = Object.assign(
+      {},
+      {
+        localThresholdMS: DEFAULT_LOCAL_THRESHOLD_MS,
```
daprahamian: Can we make a `DEFAULT_OPTIONS` obj?

lib/sdam/topology.js
```diff
@@ -0,0 +1,416 @@
...
+        serverSelectionTimeoutMS: DEFAULT_SERVER_SELECTION_TIMEOUT,
+        heartbeatFrequencyMS: DEFAULT_HEARTBEAT_FREQUENCY
+      },
+      options
+    );
+
+    const topologyType =
```
daprahamian: Nested Ternery :(

I think a function with a name like `topologyTypeFromSeedlist(seedlist, options)`? Or even just `topologyType(seedlist, options)`?

lib/sdam/topology.js
```diff
@@ -0,0 +1,416 @@
...
+      seedlist.length === 1 && !options.replicaset
+        ? TopologyType.Single
+        : options.replicaset
+          ? TopologyType.ReplicaSetNoPrimary
+          : TopologyType.Unknown;
+
+    const topologyId = globalTopologyCounter++;
```
daprahamian: Can we make a `topologyCounter()`?

lib/sdam/topology.js
```diff
@@ -0,0 +1,416 @@
...
+
+    const serverDescriptions = seedlist.reduce((result, seed) => {
+      const address = seed.port ? `${seed.host}:${seed.port}` : `${seed.host}:27017`;
```
daprahamian: Lets make an `address(...)` function.

lib/sdam/topology.js
```diff
@@ -0,0 +1,416 @@
...
+      options
+    );
+
+    selectServers(
+      this,
+      selector,
+      options.serverSelectionTimeoutMS,
+      process.hrtime(),
+      (err, servers) => {
+        if (err) return callback(err, null);
```
daprahamian: Is there a reason we needed to wrap the `callback` in this arrow function?

lib/sdam/topology.js
```diff
@@ -0,0 +1,416 @@
...
+  /**
+   * Authenticate using a specified mechanism
+   *
+   * @param {String} mechanism The auth mechanism used for authentication
+   * @param {String} db The db we are authenticating against
+   * @param {Object} options Optional settings for the authenticating mechanism
+   * @param {authResultCallback} callback A callback function
+   */
+  auth(mechanism, db, options, callback) {
+    callback(null, null);
```
daprahamian: Is this not implemented yet?

lib/sdam/topology.js
```diff
@@ -0,0 +1,416 @@
...
+  }
+
+  /**
+   * Logout from a database
+   *
+   * @param {String} db The db we are logging out from
+   * @param {authResultCallback} callback A callback function
+   */
+  logout(db, callback) {
+    callback(null, null);
```
daprahamian: same Q

lib/error.js
```diff
@@ -75,8 +75,24 @@ const MongoParseError = function(message) {
...
+/**
+ * An error signifying a timeout event
+ *
+ * @class
+ * @param {Error|string|object} message The error message
+ * @property {string} message The error message
+ * @return {MongoTimeoutError} A MongoTimeoutError instance
+ * @extends {MongoError}
+ */
+const MongoTimeoutError = function(message) {
```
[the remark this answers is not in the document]
mbroadst: yes, but I was just following existing convention in the file. I'd prefer to convert them in a separate commit wholesale

lib/sdam/server_description.js
mbroadst: It could for sure, but then we would be vulnerable to potentially changing fields in future `ismaster`s. This way we are very explicit about what we are plucking from the `ismaster`, and the definition of this type only changes when we say it changes.

lib/sdam/server_description.js
mbroadst: sure

lib/sdam/server_selectors.js
mbroadst: to keep it consistent with the `readPreferenceServerSelector` below. The idea is to eventually be able to support arbitrary user-provided selectors so I wanted to standardize on a form.

lib/sdam/server_selectors.js
mbroadst: I'm all ears for a proposed change, I was just following the spec prose

lib/sdam/server_selectors.js
mbroadst: sure

lib/sdam/server_description.js
daprahamian: in that case, could we do an 

```js
ismaster = Object.assign({ ...defaultValues }, ismaster);

this.lastWriteDate = ismaster.lastWriteDate;
...
```

lib/sdam/server_selectors.js
mbroadst: yep!

lib/sdam/topology_description.js
mbroadst: tbh that feels way less readable to me for the sake of reducing very limited (and unlikely to ever change) text duplication

lib/sdam/topology_description.js
mbroadst: I feel like this request keeps coming up in code reviews, is there a particular reason you want to do that beyond readability? This is an internal type to boot, so I'm not sure what it buys us.

lib/sdam/topology_description.js
mbroadst: not according to spec

lib/sdam/topology_description.js
mbroadst: you would never have that set, there is a spec requirement that `logicalSessionTimeoutMinutes` is `null` by default

daprahamian: So we will always have either everything in the replicaset have `logicalSessionTimeoutMinutes` or everything will not? There is never a state where some have it and some don't?

lib/sdam/server_description.js
mbroadst: sure

lib/sdam/topology.js
mbroadst: what do you mean by this? a function on the class itself? a package-local variable seems fine to me, what would we gain otherwise?

lib/sdam/topology.js
mbroadst: to be used only here?

lib/sdam/topology.js
daprahamian: I just don't like the idea of incrementing the `globalTopologyCounter` inline. I'd rather that we wrap it in a `topologyCounter` function that internally handles it.

lib/sdam/topology.js
daprahamian: yeah, just to clean it up.

lib/sdam/topology.js
mbroadst: I don't follow?

lib/sdam/topology.js
mbroadst: yes these are all stubs right now

lib/sdam/topology.js
mbroadst: yes these are all stubs right now

lib/sdam/topology_description.js
mbroadst: no, all server descriptions will _always_ have either `null` or some number value

lib/sdam/topology_description.js
mbroadst: yeah this is coming in a subsequent change

lib/sdam/topology_description.js
mbroadst: I think the function just feels long because we lack destructuring, and prettier. I just put it into a text editor and put lines prettier split into multiple lines back together and the whole thing is only 70 lines long. It's definitely kind of fat, but it's literally the core functionality of the SDAM "engine" so it's going to be long (I already refactored a ton of methods out from the base implementation). If you're saying it's unreadable that's another story, but I just gave it another glance and it seemed pretty straightforward to me

lib/sdam/topology_description.js
mbroadst: I dunno, I'm not convinced. Should we make a class called `TopologyUpdateParameters`? I think it's a little overkill just for the benefit of having named parameters. The named parameters example in that article is also not very compelling - agreed, definitely don't use "magic" boolean parameters. In this case, we are very clearly passing in named variables, essentially giving us the same benefit. On top of that, requiring an extra allocation of an object just to call a function seems needlessly wasteful.  I we can discuss POJOs for the future, but I don't think this PR is the place to make that choice

lib/sdam/topology.js
daprahamian: Ah, I missed `randomSelection`. Nvm.

PR #308, opened 2018-05-10
fix(transactions): do not send txnNumber for non-write commands

daprahamian:
We started sending txnNumber for non-transaction non-retryable
commands, which causes errors against transaction-enabled servers

Fixes NODE-1466

### DIFF
--- a/lib/topologies/mongos.js
+++ b/lib/topologies/mongos.js
@@ -895,6 +895,7 @@ var executeWriteOperation = function(self, op, ns, ops, options, callback) {
   }
 
   // increment and assign txnNumber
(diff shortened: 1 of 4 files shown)


ShaneHarvey:
Hmmm, I think this should have been caught in at least one of the transactions spec tests. I'll check and add one that confirms drivers don't send txnNumber for non-transaction commands.

For now, can you add a test to confirm your fix?

daprahamian:
@ShaneHarvey will do.

daprahamian:
@ShaneHarvey added unit tests

PR #309, opened 2018-05-21
feat(Error): adding error metadata field

daprahamian:


lib/error.js
```diff
@@ -2,6 +2,9 @@
 
 var util = require('util');
 
+const mongoContext = Symbol('mongoContext');
```
mbroadst: my thought about this was that we would have `MongoError[<symbol for context>] = {}` where the context _is_ a kitchen sink, and we don't need to protect it with symbols. Also, the symbol should be exported from this file, but not attached to the type itself

daprahamian: Can you clarify whether or not you want `MongoContext[<symbol>] = {}` or `new MongoContext()[<symbol>] = {}`? I'm not sure what the point of the first one is.

Also, I like attaching relevant symbols to the actual class, rather than just having them floating around in space.

mbroadst: The latter, i.e. if I have an instance of an error I should be able to access the [<symbol>] of that instance to check for things we attach. To my second point, I don't think it _would_ be floating around in space: it's a package variable thats exported:
```
const MongoContextSymbole = new Symbol('mongoContext');
... snip ...
module.exports = { MongoError, MongoNetworkError, MongoContextSymbol }
```

Generally this is not something we're recommending for normal user consumption, I don't think we should attach it to the `MongoError` namespace.

daprahamian: Ok, I can do that.

The reasoning behind protecting it with symbols is simply to ensure consistency and ease of debugging in the future.

```js
err[MongoContextSymbol].isGetMore = true;
...
if (err[MongoContextSymbol].isGetMore)
```

is more prone to error b/c we are expecting two different projects to check for the same named key.

mbroadst: IMO its just a little cumbersome to manage a potentially large number of exported symbols than it's worth, but I also basically care about 2/10 on this one 😄

daprahamian:
Updated.

lib/error.js
```diff
@@ -94,5 +98,6 @@ module.exports = {
   MongoError,
   MongoNetworkError,
   MongoParseError,
-  MongoTimeoutError
+  MongoTimeoutError,
+  mongoErrorContext
```
mbroadst: just to continue my reign of nitpicky terror: can we rename this to be more self explanatory? `MongoErrorContextSymbol` or something similar?

daprahamian: ok, I'll make it `mongoErrorContextSymbol`, since it is not a constructor 😉

PR #311, opened 2018-05-29
feat(retryableWrites): adding more conditions for retries

daprahamian:
Fixes NODE-1456

~WIP b/c I am working on tests to verify it works in native.~ Tests in mongodb/node-mongodb-native#1714


daprahamian:
No longer WIP

lib/isRetryableError.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
```
mbroadst: can we rename this file, or ideally merge it with `error.js`? At very least it doesn't follow filename conventions in the module (we tend to use snake_case, all lowercase) It also just feels to specific to make it its own file - this is specific to errors, I would immediately think to look in the errors definition file

daprahamian: updated to be part of `lib/error.js`

PR #314, opened 2018-06-03
NODE-1259(continued): add support for monitoring, and improve server selection

mbroadst:
NOTE: this builds on the previous PR so it will look a little huge until that is rebased.

Most notably here we introduce the ability to monitor servers, in order to keep the topology state machine up-to-date. A cool feature introduced, once monitoring was possible, is that we no longer have to explicitly call `connect` on a topology object.

This works:
```
it('should work without explicitly calling connect', function(done) {
  const config = this.configuration;
  const client = new Topology({ host: config.host, port: config.port });
  const ns = `${this.configuration.db}.insert_test`;

  client.insert(ns, { a: 1 }, function(err, result) {
    expect(err).to.not.exist;
    expect(result).to.exist;
    done();
  });
});
```

### DIFF
--- a/lib/sdam/monitoring.js
+++ b/lib/sdam/monitoring.js
@@ -1,5 +1,8 @@
 'use strict';
 
+const ServerDescription = require('./server_description').ServerDescription;
+const calculateDurationInMs = require('../utils').calculateDurationInMs;
+
 /**
  * Published when server description changes, but does NOT include changes to the RTT.
  *
@@ -88,7 +91,7 @@ class ServerHeartbeatStartedEvent {
 /**
  * Fired when the server monitor’s ismaster succeeds.
  *
- * @param {Number} duration The execution time of the event
+ * @param {Number} duration The execution time of the event in ms
  * @param {Object} reply The command reply
  * @param {Object} connectionId The connection id for the command
  */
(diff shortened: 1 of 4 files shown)


mbroadst:
@daprahamian this is ready for review finally

lib/sdam/topology.js
```diff
@@ -414,12 +420,46 @@ function selectServers(topology, selector, timeout, start, callback) {
...
-  // TODO: loop this, add monitoring
+  const retrySelection = () => {
+    // ensure all server monitors are attempting monitoring immediately
+    topology.s.servers.forEach(server => server.monitor());
+
+    const iterationTimer = setTimeout(() => {
+      callback(new MongoTimeoutError('Server selection timed out due to monitoring'));
+    }, topology.s.minHeartbeatIntervalMS);
+
+    topology.once('topologyDescriptionChanged', () => {
```
daprahamian: Is there anything here to ensure that, if the callback has already been called with an error, we won't call it here?

lib/sdam/monitoring.js
```diff
@@ -101,16 +104,104 @@ class ServerHeartbeatSucceededEvent {
...
+    server.s.monitorId = setTimeout(
+      () => monitorServer(server),
+      server.s.options.heartbeatFrequencyMS
+    );
+  };
+
+  // run the actual monitoring loop
+  server.s.monitoring = true;
+  checkServer((err, isMaster) => {
+    if (err) {
```
daprahamian: Can we switch this to `if (!err)` to reduce the indentation?

lib/sdam/topology.js
mbroadst: I've read this through a few times and I can't see how the callback would be called twice after an error has called it back. Can you show me how that might happen?

PR #316, opened 2018-06-11
chore(evergreen): add 4.0 to evergreen matrix

daprahamian:
Fixes NODE-1495

Is there a reason we are only using a 2d matrix? [It seems like C-Sharp uses a 5d one](https://github.com/mongodb/mongo-csharp-driver/blob/master/evergreen/evergreen.yml#L440).

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -269,6 +269,30 @@ tasks:
             VERSION: "3.6"
             TOPOLOGY: "sharded_cluster"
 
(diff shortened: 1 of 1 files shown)


mbroadst:
@daprahamian we use a 2d matrix because the work is incomplete to port to evergreen. Also, we don't presently use mongo-orchestration, so you can't do a 1:1 comparison between our projects yet.

PR #317, opened 2018-06-12
fix(wireprotocol): only send bypassDocumentValidation if true

shsaskin:
The bypassDocumentValidation key will only be set if it explicitly
receives a true, otherwise it remains undefined.

Fixes NODE-1492

### DIFF
--- a/test/tests/unit/wire_protocol_test.js
+++ b/test/tests/unit/wire_protocol_test.js
@@ -1,0 +1,58 @@
+'use strict';
+
+const expect = require('chai').expect;
+const bson = require('bson');
+const wireProtocol2_6 = require('../../../lib/wireprotocol/2_6_support.js');
+const wireProtocol3_2 = require('../../../lib/wireprotocol/3_2_support.js');
+
+const MockPool = {};
+
+MockPool.write = function(cmd, options, callback) {
+  callback(null, cmd);
+};
+
+describe('WireProtocol', function() {
+  it('2.6 should only set bypassDocumentValidation to true if explicitly set by user to true', function(done) {
+    const options = { bypassDocumentValidation: true };
+    wireProtocol2_6.prototype.insert(MockPool, '', '', bson, 'ops', options, function(err, data) {
+      if (err) {
+        done(err);
+      }
+      expect(data.query.bypassDocumentValidation).to.equal(true);
+      done();
+    });
+  });
+
+  it('2.6 should not set bypassDocumentValidation to true if not explicitly set by user to true', function(done) {
+    const options = { bypassDocumentValidation: false };
+    wireProtocol2_6.prototype.insert(MockPool, '', '', bson, 'ops', options, function(err, data) {
+      if (err) {
+        done(err);
+      }
+      expect(data.query.bypassDocumentValidation).to.equal(undefined);
+      done();
+    });
+  });
+
+  it('3.2 should only set bypassDocumentValidation to true if explicitly set by user to true', function(done) {
+    const options = { bypassDocumentValidation: true };
+    wireProtocol3_2.prototype.insert(MockPool, '', '', bson, 'ops', options, function(err, data) {
+      if (err) {
+        done(err);
+      }
+      expect(data.query.bypassDocumentValidation).to.equal(true);
+      done();
+    });
+  });
+
+  it('3.2 should not set bypassDocumentValidation to true if not explicitly set by user to true', function(done) {
+    const options = { bypassDocumentValidation: false };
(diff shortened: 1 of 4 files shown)


test/tests/unit/wire_protocol_test.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
+
+const expect = require('chai').expect;
+const bson = require('bson');
+const wireProtocol2_6 = require('../../../lib/wireprotocol/2_6_support.js');
+const wireProtocol3_2 = require('../../../lib/wireprotocol/3_2_support.js');
+
+const MockPool = {};
```
mbroadst: can we actually make this a `class MockPool { write() { ... } }` so we can potentially reuse it for other purposes in the future?

test/tests/unit/wire_protocol_test.js
```diff
@@ -0,0 +1,58 @@
...
+
+MockPool.write = function(cmd, options, callback) {
+  callback(null, cmd);
+};
+
+describe('WireProtocol', function() {
+  it('2.6 should only set bypassDocumentValidation to true if explicitly set by user to true', function(done) {
+    const options = { bypassDocumentValidation: true };
```
mbroadst: here you would create a `const pool = new MockPool( /* maybe future options */ )` and pass that into the `insert` method.

test/tests/unit/wire_protocol_test.js
```diff
@@ -0,0 +1,58 @@
...
+    wireProtocol2_6.prototype.insert(MockPool, '', '', bson, 'ops', options, function(err, data) {
```
mbroadst: similarly I would recommend creating an instance of the `WireProtocol`: `const wireProtocol = new WireProtocol2_6();` and using `wireProtocol.insert`

mbroadst: finally, I think for readability-sake in the future it would be best to name the variables you are passing in fake data for, e.g. 
```
const ns = 'fake.namespace';
const ismaster = {};
...

wireProtocol.insert(pool, ns, ismaster, ...)
```

test/tests/unit/wire_protocol_test.js
```diff
@@ -1,58 +1,55 @@
...
+const Pool = require('../../../lib/connection/pool.js');
 const wireProtocol2_6 = require('../../../lib/wireprotocol/2_6_support.js');
 const wireProtocol3_2 = require('../../../lib/wireprotocol/3_2_support.js');
 
-const MockPool = {};
-
-MockPool.write = function(cmd, options, callback) {
-  callback(null, cmd);
+const fake_topology = 'fake_topology';
+const fake_bson = {
```
daprahamian: You could probably replace this with `sinon.createStubInstance(bson)`.

test/tests/unit/wire_protocol_test.js
```diff
@@ -1,58 +1,55 @@
...
+  serialize: () => {},
+  deserialize: () => {}
 };
 
+const pool = new Pool(fake_topology, { bson: fake_bson });
```
daprahamian: these two lines could be `const pool = sinon.createStubInstance(Pool)`

test/tests/unit/wire_protocol_test.js
```diff
@@ -1,58 +1,55 @@
...
-      done();
-    });
+  it('3.2 should not set bypassDocumentValidation to anything if not explicitly set by user to true', function() {
+    testPoolWrite(false, new wireProtocol3_2(), undefined);
   });
+
+  function testPoolWrite(bypassDocumentValidation, wireProtocol, expected) {
+    const isMaster = {};
+    const ns = 'fake.namespace';
+    const ops = 'fake.ops';
```
daprahamian: `ops` should be an array of operations. Ideally, you would do something like:

```js
const ops = [ {a: 1}, {b: 2} ];
```

test/tests/unit/wire_protocol_test.js
```diff
@@ -1,58 +1,55 @@
...
+const sinon = require('sinon');
+const Pool = require('../../../lib/connection/pool.js');
 const wireProtocol2_6 = require('../../../lib/wireprotocol/2_6_support.js');
 const wireProtocol3_2 = require('../../../lib/wireprotocol/3_2_support.js');
 
-const MockPool = {};
-
-MockPool.write = function(cmd, options, callback) {
-  callback(null, cmd);
+const fake_topology = 'fake_topology';
```
daprahamian: The topology is normally an object. I don't think it is worth it to stub the whole topology, but it should probably be an empty object `{}` instead of a string.

PR #319, opened 2018-06-18
NODE-1374 Implement final revision of transactions (core)

mbroadst:
This is the core component to the changes required to pass the current revision of the transactions specification YAML tests.

### DIFF
--- a/lib/sessions.js
+++ b/lib/sessions.js
@@ -139,16 +139,6 @@ class ClientSession extends EventEmitter {
   }
 
   /**
-   * Increment the statement id on the internal ServerSession
-   *
-   * @param {Number} [operationCount] the number of operations performed
-   */
-  incrementStatementId(operationCount) {
-    operationCount = operationCount || 1;
-    this.serverSession.stmtId += operationCount;
-  }
-
-  /**
    * @returns whether this session is current in a transaction or not
    */
   inTransaction() {
@@ -170,7 +160,6 @@ class ClientSession extends EventEmitter {
 
     // increment txnNumber and reset stmtId to zero.
     this.serverSession.txnNumber += 1;
-    this.serverSession.stmtId = 0;
 
     // set transaction options, we will use this to determine if we are in a transaction
     this.transactionOptions = Object.assign({}, options || this.defaultTransactionOptions);
--- a/lib/sessions.js
+++ b/lib/sessions.js
@@ -7,6 +7,8 @@ const Binary = BSON.Binary;
 const uuidV4 = require('./utils').uuidV4;
 const MongoError = require('./error').MongoError;
 const isRetryableError = require('././error').isRetryableError;
+const MongoNetworkError = require('./error').MongoNetworkError;
+const MongoWriteConcernError = require('./error').MongoWriteConcernError;
 
 function assertAlive(session, callback) {
   if (session.serverSession == null) {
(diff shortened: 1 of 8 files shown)


lib/sessions.js
```diff
@@ -22,6 +24,94 @@ function assertAlive(session, callback) {
   return true;
 }
 
+// Client session states
+const TxnState = {
```
daprahamian: Can we refactor this into something that uses an IIFE to make it less verbose?

I was thinking something like:

```js
let TxnState;
let txnStateTransition;
(() => {
  const NO_TRANSACTION = 'NO_TRANSACTION';
  const STARTING_TRANSACTION = 'STARTING_TRANSACTION';
  const TRANSACTION_IN_PROGRESS: 'TRANSACTION_IN_PROGRESS';
  const TRANSACTION_COMMITTED: 'TRANSACTION_COMMITTED';
  const TRANSACTION_COMMITTED_EMPTY: 'TRANSACTION_COMMITTED_EMPTY';
  const TRANSACTION_ABORTED: 'TRANSACTION_ABORTED';

  const stateMachine = {...};

  TxnState = { NO_TRANSACTION, STARTING_TRANSACTION, TRANSACTIOKN_IN_PROGRESS ... };
  txnStateTransition = function(txnState, nextState) {
  };
})();
```

daprahamian: Also, this is adding 114 lines to this file. Can we put it in a separate file?

lib/sessions.js
```diff
@@ -219,73 +308,111 @@ class ClientSession extends EventEmitter {
...
-      callback(new MongoError('No transaction started'));
+  // handle any initial problematic cases
+  let txnState = session.transaction.state;
+
+  if (txnState === TxnState.NO_TRANSACTION) {
+    callback(new MongoError('No transaction started'));
+    return;
+  }
+
+  if (commandName === 'commitTransaction') {
```
daprahamian: Is this a `const` somewhere?

lib/wireprotocol/3_2_support.js
```diff
@@ -346,15 +368,17 @@ WireProtocol.prototype.command = function(bson, ns, cmd, cursorState, topology,
   } else if (cmd) {
     query = setupCommand(bson, ns, cmd, cursorState, topology, options);
   } else {
-    throw new MongoError(f('command %s does not return a cursor', JSON.stringify(cmd)));
+    return new MongoError(f('command %s does not return a cursor', JSON.stringify(cmd)));
```
daprahamian: Can we use template strings here?

lib/sessions.js
```diff
@@ -219,73 +308,111 @@ class ClientSession extends EventEmitter {
...
-        return clientSession.topology.command(
-          'admin.$cmd',
-          command,
-          { session: clientSession },
-          (_err, _reply) => commandHandler(transactionError(_err), _reply)
-        );
-      }
-
-      commandHandler(transactionError(err), reply);
+  session.topology.command('admin.$cmd', command, { session }, (err, reply) => {
```
daprahamian: This makes me feel weird for some reason, but I can't figure out a good way to refactor this.

I'd love for there to be some way to not write the `session.topology.command` out more than once.

mbroadst: I'm going to leave this be, but certainly submit a PR if you come up with something you think is better!

PR #321, opened 2018-06-19
test(atlas): atlas connectivity tests

daprahamian:
Fixes NODE-1221

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -136,6 +136,15 @@ functions:
           ${PREPARE_SHELL}
           NODE_LTS_NAME=${NODE_LTS_NAME} sh ${PROJECT_DIRECTORY}/.evergreen/install-dependencies.sh
 
+  "run atlas tests":
+    - command: shell.exec
+      params:
+        working_dir: "src"
+        silent: true
+        script: |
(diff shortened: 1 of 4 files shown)


atlas.js
```diff
@@ -0,0 +1,120 @@
...
+    .then(() => topologyIsMaster(topology))
+    .then(() => topologyFindCmd(topology))
+    .then(() => console.log(`${config.name} passed`))
+    .catch(e => {
+      console.log(`${config.name} failed`);
+      throw e;
+    });
+}
+
+configs
```
mbroadst: is there a particular reason you chose not to use mocha for this?

daprahamian: a) it seemed like overkill
b) I explicitly wanted to not have a possibility of an error message leaking out so as to not expose the url

mbroadst: My concern is that this special cases a class of test we have, how is "atlas connect" different than "ssl" or "auth"?

daprahamian: a. It's different in that we don't run `ssl` or `auth` anymore ;)
b. The explicit requirements that we not expose the atlas urls has me a little paranoid.

I kind of view this specific test case as similar to driverBench.

mbroadst: 👍

PR #322, opened 2018-06-20
fix(error): attach command response to MongoWriteConcernError

kvwalker:
Fixes NODE-1521

### DIFF
--- a/lib/connection/pool.js
+++ b/lib/connection/pool.js
@@ -581,6 +581,13 @@ function messageHandler(self) {
           }
 
           if (responseDoc.writeConcernError) {
+            if (responseDoc.ok === 1) {
+              return handleOperationCallback(
+                self,
+                workItem.cb,
+                new MongoWriteConcernError(responseDoc.writeConcernError, responseDoc)
+              );
+            }
             return handleOperationCallback(
               self,
               workItem.cb,
--- a/lib/connection/pool.js
+++ b/lib/connection/pool.js
@@ -581,18 +581,11 @@ function messageHandler(self) {
           }
 
           if (responseDoc.writeConcernError) {
-            if (responseDoc.ok === 1) {
-              return handleOperationCallback(
-                self,
-                workItem.cb,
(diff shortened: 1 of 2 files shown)


lib/error.js
```diff
@@ -131,12 +131,17 @@ function isRetryableError(error) {
...
  * @param {Error|string|object} message The error message
  * @property {string} message The error message
+ * @property {object} result The result document (provided if ok: 1)
  * @return {MongoWriteConcernError} A MongoWriteConcernError instance
  * @extends {MongoError}
  */
-const MongoWriteConcernError = function(message) {
-  MongoError.call(this, message);
+const MongoWriteConcernError = function(message, result) {
+  MongoError.call(this, message, result);
```
daprahamian: Why do we need to pass `result` to `MongoError`? It doesn't appear to do anything with it.

lib/connection/pool.js
```diff
@@ -581,6 +581,13 @@ function messageHandler(self) {
           }
 
           if (responseDoc.writeConcernError) {
+            if (responseDoc.ok === 1) {
```
daprahamian: What about doing a ternary to create the error, and then using a single callback?

```js
if (responseDoc.writeConcernError) {
  const err = responseDoc.ok === 1 ?
    new MongoWriteConcernError(responseDoc.writeConcernError, responseDoc) :
    new MongoWriteConcernError(responseDoc.writeConcernError);
  return handleOperationCallback(self, workItem.cb, err);
}
```

lib/error.js
kvwalker: oops, thanks for catching that!

lib/error.js
```diff
@@ -131,12 +131,17 @@ function isRetryableError(error) {
  * @class
  * @param {Error|string|object} message The error message
  * @property {string} message The error message
+ * @property {object} result The result document (provided if ok: 1)
```
mbroadst: Please wrap the name of the argument with brackets to indicate it is optional, and add it to the `@param` documentation.

PR #323, opened 2018-06-22
feat(kerberos): expose warning for kerberos mismatch versions

daprahamian:
We were originally blindly catching errors for a kerberos version
mismatch, and silently failing. This exposes the error message
to users if it exists.

Fixes NODE-1523


lib/auth/getKerberos.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const requireOptional = require('require_optional');
```
mbroadst: can't we just put this into `lib/utils` similar to `retrieveBSON`? I'm starting to see the possible explosive growth of tiny files on its way 😄

lib/auth/getKerberos.js
```diff
@@ -0,0 +1,18 @@
...
+let MongoAuthProcess = null;
+
+try {
+  const kerberos = requireOptional('kerberos');
+  if (kerberos) {
+    Kerberos = kerberos.Kerberos;
+    MongoAuthProcess = kerberos.processes.MongoAuthProcess;
+  }
+} catch (err) {
+  console.warn(err.message);
```
mbroadst: I think the reason this was ignored in the first place is because it is meant to be optional. Rather, I think the warning should be suppressed if the module isn't found, but it should _rethrow_ for any other error

PR #325, opened 2018-06-25
fix(sessions): add `toBSON` method to `ClientSession`

kvwalker:
Fixes [NODE-1518](https://jira.mongodb.org/browse/NODE-1518)

### DIFF
--- a/lib/sessions.js
+++ b/lib/sessions.js
@@ -220,6 +220,10 @@ class ClientSession extends EventEmitter {
       );
     });
   }
+
+  toBSON() {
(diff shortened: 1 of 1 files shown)


lib/sessions.js
```diff
@@ -220,6 +220,10 @@ class ClientSession extends EventEmitter {
       );
     });
   }
+
+  toBSON() {
+    throw new Error('ClientSession cannot be serialized.');
```
daprahamian: Can we make the message `ClientSession cannot be serialized to BSON`?

PR #326, opened 2018-06-29
fix(commands): check doc.cursor errors

kvwalker:
Fixes [NODE-1528](https://jira.mongodb.org/browse/NODE-1528)

### DIFF
--- a/lib/connection/commands.js
+++ b/lib/connection/commands.js
@@ -484,8 +484,15 @@ Response.prototype.parse = function(options) {
     // Deserialize but keep the array of documents in non-parsed form
     var doc = this.bson.deserialize(document, _options);
 
+    if (doc instanceof Error) {
+      throw doc;
+    }
     // Get the documents
-    this.documents = doc.cursor[documentsReturnedIn];
+    if (doc.cursor) {
+      this.documents = doc.cursor[documentsReturnedIn];
+    } else {
+      throw new Error('Cursor not found');
+    }
     this.numberReturned = this.documents.length;
     // Ensure we have a Long valie cursor id
     this.cursorId =
--- a/lib/connection/commands.js
+++ b/lib/connection/commands.js
@@ -3,6 +3,7 @@
 var retrieveBSON = require('./utils').retrieveBSON;
 var BSON = retrieveBSON();
 var Long = BSON.Long;
+const MongoError = require('../error').MongoError;
 
 // Incrementing request id
 var _requestId = 0;
@@ -491,7 +492,7 @@ Response.prototype.parse = function(options) {
     if (doc.cursor) {
       this.documents = doc.cursor[documentsReturnedIn];
     } else {
-      throw new Error('Cursor not found');
+      throw new MongoError('Cursor not found');
     }
     this.numberReturned = this.documents.length;
     // Ensure we have a Long valie cursor id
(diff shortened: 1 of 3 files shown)


lib/connection/commands.js
```diff
@@ -484,8 +484,15 @@ Response.prototype.parse = function(options) {
...
 
+    if (doc instanceof Error) {
+      throw doc;
+    }
     // Get the documents
-    this.documents = doc.cursor[documentsReturnedIn];
+    if (doc.cursor) {
+      this.documents = doc.cursor[documentsReturnedIn];
+    } else {
+      throw new Error('Cursor not found');
```
daprahamian: This case is usually hit when we get a server error. Can we throw that error instead of a simple `Error`?

lib/connection/commands.js
```diff
@@ -484,8 +485,15 @@ Response.prototype.parse = function(options) {
...
 
+    if (doc instanceof Error) {
+      throw doc;
+    }
     // Get the documents
-    this.documents = doc.cursor[documentsReturnedIn];
+    if (doc.cursor) {
+      this.documents = doc.cursor[documentsReturnedIn];
+    } else {
+      throw new MongoError('Cursor not found');
```
herbrandson: I see. This should handle the case of `doc.ok === 0`. But, perhaps you could still use `errmsg`? Maybe something like ` throw new MongoError(doc.errmsg || 'Cursor not found')`?

daprahamian: I agree with this. We could do `throw new MongoError(doc.errmsg ? doc : 'Cursor not found');`, so we could catch any error codes on the doc.

test/tests/unit/response_test.js
```diff
@@ -0,0 +1,89 @@
...
+      test.server.setMessageHandler(request => {
+        const doc = request.document;
+        if (doc.ismaster) {
+          request.reply(
+            Object.assign({}, mock.DEFAULT_ISMASTER, {
+              maxWireVersion: 6
+            })
+          );
+        } else if (doc.find) {
+          commands.push(doc);
```
mbroadst: no need to push commands onto that stack if you'll never be checking it for expectations

test/tests/unit/response_test.js
```diff
@@ -0,0 +1,89 @@
...
+          request.reply({
+            cursor: {
+              id: Long.fromNumber(1),
+              ns: 'test.test',
+              firstBatch: []
+            },
+            ok: 1
+          });
+        } else if (doc.getMore) {
+          console.log('getMore');
```
mbroadst: I think you left this in by accident

test/tests/unit/response_test.js
```diff
@@ -0,0 +1,89 @@
...
+          commands.push(doc);
+          request.reply(errdoc);
+        }
+      });
+
+      client.on('error', done);
+      client.once('connect', () => {
+        const cursor = client.cursor(
```
mbroadst: I think maybe we can just do: `const cursor = client.cursor('test.test', { find: 'test' });`

test/tests/unit/response_test.js
```diff
@@ -0,0 +1,89 @@
...
+    return mock.createServer().then(mockServer => {
+      test.server = mockServer;
+    });
+  });
+
+  it('should throw when document is error', {
+    metadata: { requires: { topology: ['single'] } },
+    test: function(done) {
+      const errdoc = {
+        ok: 0,
```
mbroadst: It's nice to have the whole error response, but I think generally you _just_ need `errmsg` to trigger the code you're looking to trigger right? We should try to be least verbose in these types of mock tests IMO

PR #327, opened 2018-06-30
Do not force keys to lower case in URL parser

astanciu:
Addresses issue in NODE-1542.

The URL Parser sets all keys to lowercase, which causes something like this:
`...?replicaSet=rs` to become `replicaset: rs`, effectively making any camelCased option undefined.


astanciu:
Closing this, all the failing tests indicate this isn't the right approach; other things expect the keys to be lower cased.

PR #328, opened 2018-07-09
docs(sessions): updating documentation for sessions

daprahamian:
Major update to sessions documentation. We are exposing the
`ClientSession` class, as well as Transaction options. Also removes
Disqus from the core docs.


lib/sessions.js
```diff
@@ -149,7 +166,7 @@ class ClientSession extends EventEmitter {
   }
 
   /**
-   * @returns whether this session is current in a transaction or not
+   * @returns {boolean} whether this session is current in a transaction or not
```
kvwalker: Since we're fixing things here, should 'current' be 'currently' instead?

lib/topologies/read_preference.js
```diff
@@ -1,9 +1,18 @@
...
+ * @class
+ * @param {string} mode A string describing the read preference mode (primary|primaryPreferred|secondary|secondaryPreferred|nearest)
+ * @param {array} tags The tags object
+ * @param {object} [options] Additional read preference options
+ * @param {number} [options.maxStalenessSeconds] Max secondary read staleness in seconds, Minimum value is 90 seconds.
+ * @property {string} mode The read preference mode (primary|primaryPreferred|secondary|secondaryPreferred|nearest)
+ * @property {array} tags The tags object
+ * @property {object} options Additional read preference options
+ * @property {number} maxStalenessSeconds MaxStalenessSeconds value for the read preference
+ * @return {ReadPreference}
```
kvwalker: It looks like these are repeated so maybe clean up the docblock above?

lib/topologies/read_preference.js
```diff
@@ -147,7 +144,8 @@ ReadPreference.prototype.slaveOk = function() {
 /**
  * Are the two read preference equal
  * @method
- * @return {boolean}
+ * @param {ReadPreference} readPreference the other read preference
```
kvwalker: Instead of "the other read preference", I think "The read preference with which to check equality" would be more clear.

lib/transactions.js
```diff
@@ -46,16 +46,43 @@ let stateMachine;
...
+ * @property {'local'|'available'|'majority'|'linearizable'|'snapshot'} level The readConcern Level
+ * @see https://docs.mongodb.com/manual/reference/read-concern/
+ */
+
+/**
+ * A MongoDB WriteConcern, which describes the level of acknowledgement
+ * requested from MongoDB for write operations.
+ * @typedef {Object} WriteConcern
+ * @property {number|'majority'|string} [w=1] requests acknowledgement that the write operation has
+ * propagated to a specified number of mongod
```
kvwalker: "mongod" => "mongod hosts"?

PR #329, opened 2018-07-12
refactor(collation): utility function for checking collation support

rweinberger:
Created a utility function collationNotSupported() that will check
if collation is supported by the server, and if not, call the callback 
with an error.

Fixes NODE-1522

### DIFF
--- a/lib/cursor.js
+++ b/lib/cursor.js
@@ -635,8 +635,10 @@ var nextFunction = function(self, callback) {
     // Set as init
     self.cursorState.init = true;
 
-    // if collation not supported, will call callback with new MongoError
-    if (collationNotSupported(self.server, self.cmd, callback)) return;
+    // error if collation not supported
+    if (collationNotSupported(self.server, self.cmd)) {
+      return callback(new MongoError(`server ${self.server.name} does not support collation`));
+    }
 
     try {
       self.query = self.server.wireProtocolHandler.command(
--- a/lib/cursor.js
+++ b/lib/cursor.js
@@ -6,6 +6,7 @@ const MongoError = require('./error').MongoError;
 const MongoNetworkError = require('./error').MongoNetworkError;
 const mongoErrorContextSymbol = require('./error').mongoErrorContextSymbol;
 const f = require('util').format;
+const collationNotSupported = require('./utils').collationNotSupported;
 
 var BSON = retrieveBSON(),
   Long = BSON.Long;
@@ -634,10 +635,8 @@ var nextFunction = function(self, callback) {
     // Set as init
     self.cursorState.init = true;
 
-    // Server does not support server
-    if (self.cmd && self.cmd.collation && self.server.ismaster.maxWireVersion < 5) {
-      return callback(new MongoError(f('server %s does not support collation', self.server.name)));
-    }
+    // if collation not supported, will call callback with new MongoError
+    if (collationNotSupported(self.server, self.cmd, callback)) return;
 
     try {
       self.query = self.server.wireProtocolHandler.command(
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -2,7 +2,6 @@
 
 const crypto = require('crypto');
 const requireOptional = require('require_optional');
-const MongoError = require('./error').MongoError;
 
 /**
  * Generate a UUIDv4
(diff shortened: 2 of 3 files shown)


lib/utils.js
```diff
@@ -48,10 +50,26 @@ try {
...
+ * Checks that collation is supported by server.
+ *
+ * @param {Server} [server] to check against
+ * @param {object} [cmd] object where collation may be specified
+ * @param {function} [callback] callback function
+ * @return true if server does not support collation
+ */
+const collationNotSupported = (server, cmd, callback) => {
+  if (cmd && cmd.collation && server.ismaster && server.ismaster.maxWireVersion < 5) {
+    callback(new MongoError(f('server %s does not support collation', server.name)));
```
kvwalker: Could you rewrite this to avoid the use of `f`?

lib/utils.js
```diff
@@ -48,10 +50,26 @@ try {
...
 
+/*
+ * Checks that collation is supported by server.
+ *
+ * @param {Server} [server] to check against
+ * @param {object} [cmd] object where collation may be specified
+ * @param {function} [callback] callback function
+ * @return true if server does not support collation
+ */
+const collationNotSupported = (server, cmd, callback) => {
```
daprahamian: since this is a standalone function, I'd prefer to use a function definition:

```js
function collationNotSupported(server, cmd, callback) {
```

lib/utils.js
```diff
@@ -48,10 +50,26 @@ try {
...
+  if (cmd && cmd.collation && server.ismaster && server.ismaster.maxWireVersion < 5) {
+    callback(new MongoError(f('server %s does not support collation', server.name)));
+    return true;
```
daprahamian: I can't thing of a better solution, but I'd like to have a discussion about invoking the callback and returning a boolean to help the invokee return early. CC @mbroadst @kvwalker

mbroadst: I'm conflicted as well. Removing returning the `MongoError` at the actual call site makes it feel like we are decreasing readability: on reading it's not immediately clear that `callback` is ever called, though it is passed somewhere.

mbroadst:
Question: shouldn't we move this to `mongodb-core` completely?

lib/utils.js
rweinberger: Alternatively, could we just have the `collationNotSupported` function just return the boolean indicating whether collation is supported? Then the code might look like: 
`if collationNotSupported(server, cmd) return callback (new MongoError('...'));`
Although I guess that still leaves repetitive code

daprahamian:
@mbroadst the plan was to move it to `mongodb-core` entirely once we resolved the design issue [here](https://github.com/mongodb-js/mongodb-core/pull/329#discussion_r202473565).

lib/utils.js
mbroadst: I think my preference is to leave the callbacks in place, and reduce this function to determining if collation is supported or not. It gives us more control over what message is returned, and whether we want to throw or return an async error. I also think it improves readability. I'm about a 4/10 on this, but let's see what y'all think.

lib/cursor.js
```diff
@@ -634,10 +635,8 @@ var nextFunction = function(self, callback) {
     // Set as init
     self.cursorState.init = true;
 
-    // Server does not support server
-    if (self.cmd && self.cmd.collation && self.server.ismaster.maxWireVersion < 5) {
-      return callback(new MongoError(f('server %s does not support collation', self.server.name)));
-    }
+    // if collation not supported, will call callback with new MongoError
+    if (collationNotSupported(self.server, self.cmd, callback)) return;
```
mbroadst: @rweinberger I thought the idea was to call the callback in place, instead of passing it to the helper function:
```
    if (collationNotSupported(self.server, self.cmd)) {
      return callback(new MongoError(f('server %s does not support collation', self.server.name)));
    }
```

rweinberger: Right, I haven't actually updated it from the original design yet since I was waiting for more input into the design issue, but if this is design good with everyone I can go ahead and change it @daprahamian @kvwalker

daprahamian: I am on board with the changes. 👍

kvwalker: Me too!

PR #331, opened 2018-07-19
fix(read-preference): correct server sort for `nearest` selection

mbroadst:
The server sorting during `nearest` server selection was incorrectly
using a greater-than sign, instead of following the rules of array
sort comparitors. This fixes the issue for `pickNearest` as well
as `pickNearestMaxStalenessSeconds`, and contributes tests for
both paths.

NODE-1577


test/tests/unit/replset/read_preference_tests.js
```diff
@@ -48,4 +50,47 @@ describe('Secondaries (ReplSet)', function() {
       replSet.connect({ readPreference: new ReadPreference('secondary') });
     }
   });
+
+  it('should correctly sort servers by `lastIsMasterMS` during nearest selection', function() {
+    const state = new ReplSetState();
+    let sampleData = [
```
daprahamian: nit: `const`

Issue #292, opened 2018-04-11
connection increase problem


tiger-cub:
I set the connection pool size to 5 and ran the application.
Restarting db will continue to increase the number of connections to mongo db up to unlimit..
i found 8000 connection count when application was down.
Most connections are in keepalive state.

set the version 3.0.1 to 3.0.6..... make problem
but use 2.1.19 cannot make problem. what's different?

yg-dev:
@menme95 no! 2.1.19 has no problem. so i using that now.

menme95:
I've test the version 5.2.9 and it has been solved, if you want to update a newer version.

PR #332, opened 2018-07-22
fix(pool): don't try to reconnect if initial connection fails

vkarpov15:
Automattic/mongoose#6670 seems to be another instance of the issue I mentioned in #275: if initial connection fails, the driver will keep trying to reconnect in the background. So users that retry `MongoClient.connect()` when initial connection fails get a bunch of zombie database connections (one might call them "dead pools" :) ) if one of their reconnect tries succeeds. For example, try running this code against a standalone server that isn't running, and then start it after 10 seconds, you'll see 11 connections rather than 1.

```javascript
var mongodb = require('mongodb').MongoClient;

var url = 'mongodb://localhost:27017/edata';
var options = {
        useNewUrlParser: true,
        autoReconnect: true,
        poolSize: 3,
        reconnectTries: Number.MAX_VALUE,
        reconnectInterval: 1000,
        socketTimeoutMS: 5000
}
var conectDate = new Date();
var count = 0;


// Connect to the db
var connectWithRetry = function () {
        mongodb.connect(url, options, function (err, db) {

                
                if (!err) {
                        console.log("We are connected");
                        db.on('close', function () {
                                console.log('Close + ' + count);

                        })
                        db.on('reconnect', function () {
                                console.log('Reconnect')
                        })
                }
                else {
                        setTimeout(connectWithRetry, 1000);
                        count++;
                        console.log(conectDate + ' - Database not connected - ' + count);
                }
        });
}

connectWithRetry();
```

Just wanted to put this issue out there and suggest a potential fix. We need a consistent explanation for how to handle initial connection failures: either the driver will keep trying to reconnect, or the user should try to reconnect on their own. Right now we generally recommend the latter, but the code does a little bit of both the latter and the former.

### DIFF
--- a/lib/connection/pool.js
+++ b/lib/connection/pool.js
@@ -273,7 +273,7 @@ function reauthenticate(pool, connection, cb) {
   authenticateAgainstProvider(pool, connection, Object.keys(pool.authProviders), cb);
 }
 
-function connectionFailureHandler(self, event) {
+function connectionFailureHandler(self, event, skipReconnect) {
   return function(err) {
     if (this._connectionFailHandled) return;
     this._connectionFailHandled = true;
@@ -317,7 +317,7 @@ function connectionFailureHandler(self, event) {
     }
 
     // Start reconnection attempts
-    if (!self.reconnectId && self.options.reconnect) {
+    if (!self.reconnectId && self.options.reconnect && !skipReconnect) {
       self.reconnectId = setTimeout(attemptReconnect(self), self.options.reconnectInterval);
     }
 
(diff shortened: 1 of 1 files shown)


kereslas:
Probably same issue from #292.

daprahamian:
Actually, just dove into this: the problem is not the pool trying to reconnect, but the fact that we never destroy the pool when a mongo client fails to do an initial connect.

I will work on a fix for native and get that up ASAP

daprahamian:
@vkarpov15 I have PRs up at #335 and mongodb/node-mongodb-native#1797 to fix this.

vkarpov15:
Excellent, thanks @daprahamian . The fact that we still continue to reconnect even after initial connection fails is exactly the problem, thanks for fixing that. Any reason to keep this PR open or can I close it?

daprahamian:
@vkarpov15 we can close this.

vkarpov15:
Done

PR #336, opened 2018-08-02
(WIP) feat(connection): fast fallback

shsaskin:
Implements fast fallback with the happy eyeballs algorithm to avoid waiting for Ipv6 to error before connecting to IPv4.

Fixes Node-1580

### DIFF
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -599,13 +599,6 @@ function makeUnsecureConnection(self, family) {
   connection.setTimeout(self.connectionTimeout);
   connection.setNoDelay(self.noDelay);
 
-  connection.once('connect', function() {
-    // Set socket timeout instead of connection timeout
-    connection.setTimeout(self.socketTimeout);
-    // Emit connect event
-    // self.emit('connect', self);
-  });
-
   return connection;
 }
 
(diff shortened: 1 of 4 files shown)


test/tests/functional/rs_mocks/add_remove_tests.js
```diff
@@ -603,6 +603,7 @@ describe('ReplSet Add Remove (mocks)', function() {
             done();
           }, 1000);
         }, 500);
+        // }, 650);
```
shsaskin: The add_remove_tests that are failing pass when I change the timeout to 650ms. My current implementation is slowing down the replset connections. Is there something in my implementation that could be slowing this down?

lib/connection/connection.js
```diff
@@ -591,34 +594,87 @@ function makeUnsecureConnection(self, family) {
...
+    self.ipv6Connection = makeUnsecureConnection(self, 6);
+    setupConnection(self, self.ipv6Connection, _errorHandlerForIPv6);
+
+    self.ipv4Connection = makeUnsecureConnection(self, 4);
+    setupConnection(self, self.ipv4Connection, _errorHandler);
+
+    self.ipv6Connection.once('connect', function() {
+      if (!connected) {
+        self.connection = self.ipv6Connection;
+        onConnect(self);
```
shsaskin: I originally ended the ipv4 connection and removed the listeners here, but this was causing a timeout test to fail because the connection timeout event would not be emitted. However, now I do not know where to shutdown the connection and remove the listeners.

lib/connection/connection.js
```diff
@@ -591,34 +594,87 @@ function makeUnsecureConnection(self, family) {
...
+        connected = true;
+      }
+    });
+    self.ipv4Connection.once('connect', function() {
+      setTimeout(function() {
+        if (!connected) {
```
shsaskin: This was originally where I ended the ipv6 connection and removed the listeners, but I ran into the same problem as above.

PR #341, opened 2018-08-06
fix(uri-parser): Parse comma-separated option values

rweinberger:
Enable parsing of comma-separated option values into arrays.

Fixes NODE-1612


rweinberger:
@kvwalker I submitted a PR to the specifications repo to add a test for this in the connection string spec tests

PR #343, opened 2018-08-13
feat(connection): Implement fast fallback

shsaskin:
Implements fast fallback and refactors connection
to allow family to be specified for ssl and unsecure
connections.

Fixes NODE-1580

### DIFF
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -576,9 +576,9 @@ function prepareConnectionOptions(conn, _options) {
 }
 
 function makeConnection(conn, options, callback) {
-  const module = options.ssl ? tls : net;
+  const netModule = options.ssl ? tls : net;
 
-  const connection = module.connect(options, function() {
+  const connection = netModule.connect(options, function() {
     if (conn.ssl) {
       // Error on auth or skip
       if (connection.authorizationError && conn.rejectUnauthorized) {
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -629,7 +629,7 @@ function fastFallbackConnect(conn, _options, callback) {
   const connectionHandler = (err, _connection) => {
     if (_connection) {
       if (connection) {
-        ['error', 'timeout', 'close'].forEach(event => _connection.removeAllListeners(event));
+        _connection.removeAllListeners('error');
         _connection.unref();
         return;
       }
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -528,97 +528,142 @@ function merge(options1, options2) {
   }
 }
 
-function makeSSLConnection(self, _options) {
-  let sslOptions = {
-    socket: self.connection,
-    rejectUnauthorized: self.rejectUnauthorized
-  };
-
-  // Merge in options
-  merge(sslOptions, self.options);
-  merge(sslOptions, _options);
-
-  // Set options for ssl
-  if (self.ca) sslOptions.ca = self.ca;
-  if (self.crl) sslOptions.crl = self.crl;
-  if (self.cert) sslOptions.cert = self.cert;
-  if (self.key) sslOptions.key = self.key;
-  if (self.passphrase) sslOptions.passphrase = self.passphrase;
-
-  // Override checkServerIdentity behavior
-  if (self.checkServerIdentity === false) {
-    // Skip the identiy check by retuning undefined as per node documents
-    // https://nodejs.org/api/tls.html#tls_tls_connect_options_callback
-    sslOptions.checkServerIdentity = function() {
-      return undefined;
+function prepareConnectionOptions(self, _options) {
+  let options;
+  if (self.ssl) {
+    options = {
+      socket: self.connection,
+      rejectUnauthorized: self.rejectUnauthorized
     };
-  } else if (typeof self.checkServerIdentity === 'function') {
-    sslOptions.checkServerIdentity = self.checkServerIdentity;
-  }
 
-  // Set default sni servername to be the same as host
-  if (sslOptions.servername == null) {
-    sslOptions.servername = self.host;
-  }
+    // Merge in options
+    merge(options, self.options);
+    merge(options, _options);
 
-  // Attempt SSL connection
-  const connection = tls.connect(self.port, self.host, sslOptions, function() {
-    // Error on auth or skip
-    if (connection.authorizationError && self.rejectUnauthorized) {
-      return self.emit('error', connection.authorizationError, self, { ssl: true });
+    // Set options for ssl
+    if (self.ca) options.ca = self.ca;
+    if (self.crl) options.crl = self.crl;
+    if (self.cert) options.cert = self.cert;
+    if (self.key) options.key = self.key;
+    if (self.passphrase) options.passphrase = self.passphrase;
+
+    // Override checkServerIdentity behavior
+    if (self.checkServerIdentity === false) {
+      // Skip the identiy check by retuning undefined as per node documents
+      // https://nodejs.org/api/tls.html#tls_tls_connect_options_callback
+      options.checkServerIdentity = function() {
+        return undefined;
+      };
+    } else if (typeof self.checkServerIdentity === 'function') {
+      options.checkServerIdentity = self.checkServerIdentity;
     }
 
-    // Set socket timeout instead of connection timeout
-    connection.setTimeout(self.socketTimeout);
-    // We are done emit connect
-    self.emit('connect', self);
-  });
+    // Set default sni servername to be the same as host
+    if (options.servername == null) {
+      options.servername = self.host;
+    }
 
-  // Set the options for the connection
-  connection.setKeepAlive(self.keepAlive, self.keepAliveInitialDelay);
-  connection.setTimeout(self.connectionTimeout);
-  connection.setNoDelay(self.noDelay);
+    options = Object.assign({}, options, { host: self.host, port: self.port });
 
-  return connection;
-}
-
-function makeUnsecureConnection(self, family) {
-  // Create new connection instance
-  let connection_options;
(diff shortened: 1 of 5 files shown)


shsaskin:
The current state of the code passes most tests, but it fails the connection_tests.js by re-running the ipv6 connect with no family test and failing the second time. I also still have the timeout extended for the add_remove_tests.js to get the replset connections to pass the tests. I am currently looking into these problems, but I wanted to check if I am still going in the desired direction.

lib/connection/connection.js
```diff
@@ -528,97 +528,142 @@ function merge(options1, options2) {
...
+    options = Object.assign({}, options, { host: self.host, port: self.port });
+
+    return options;
+  } else {
+    if (self.domainSocket) {
+      options = { path: self.host };
+    } else {
+      options = { port: self.port, host: self.host };
+    }
+    return options;
```
mbroadst: move duplicated `return options`; below outside of the top level if statement

lib/connection/connection.js
```diff
@@ -528,97 +528,142 @@ function merge(options1, options2) {
...
-  if (self.key) sslOptions.key = self.key;
-  if (self.passphrase) sslOptions.passphrase = self.passphrase;
-
-  // Override checkServerIdentity behavior
-  if (self.checkServerIdentity === false) {
-    // Skip the identiy check by retuning undefined as per node documents
-    // https://nodejs.org/api/tls.html#tls_tls_connect_options_callback
-    sslOptions.checkServerIdentity = function() {
-      return undefined;
+function prepareConnectionOptions(self, _options) {
```
mbroadst: let's rename `self` to something more informative

lib/connection/connection.js
```diff
@@ -528,97 +528,142 @@ function merge(options1, options2) {
...
+    return callback(null, connection);
+  };
 
-  // Add handlers for events
-  self.connection.once('error', _errorHandler);
-  self.connection.once('timeout', timeoutHandler(self));
-  self.connection.once('close', closeHandler(self));
-  self.connection.on('data', dataHandler(self));
+  makeConnection(self, Object.assign({}, { family: 6 }, options), connectionHandler);
+  setTimeout(() => {
```
mbroadst: why is this `setTimeout` here?

lib/connection/connection.js
```diff
@@ -528,97 +528,142 @@ function merge(options1, options2) {
...
-    connection.setTimeout(self.socketTimeout);
-    // Emit connect event
-    self.emit('connect', self);
-  });
+function fastFallback(self, _options, callback) {
+  const options = prepareConnectionOptions(self, _options);
+
+  let connection;
+  const connectionHandler = (err, _connection) => {
+    const _errorHandler = errorHandler(self);
```
mbroadst: let's move the debouncing to the first thing done here. If a connection has already been established (i.e. `connection != null`), then we don't want to do _anything_ other than destroy the other established connection. We don't even care about errors.

lib/connection/connection.js
```diff
@@ -528,97 +528,142 @@ function merge(options1, options2) {
...
+          return _closeHandler(err.err);
+        case 'data':
+          return _dataHandler(err.err);
+      }
+    }
 
-  return connection;
-}
+    if (connection) {
+      _connection.removeAllListeners('error');
```
mbroadst: this can be simplified to `['error', 'timeout', 'close', 'data'].forEach(event => _connection.removeAllListeners(event))`

lib/connection/connection.js
```diff
@@ -528,97 +528,142 @@ function merge(options1, options2) {
...
+      _connection.removeAllListeners('timeout');
+      _connection.removeAllListeners('close');
+      _connection.removeAllListeners('data');
+      _connection.end();
+      _connection.unref();
```
mbroadst: you need to remember to `return` after the cleanup has happened, otherwise you will call the callback below with the selected connection twice

lib/connection/connection.js
```diff
@@ -528,97 +528,142 @@ function merge(options1, options2) {
...
   });
 
   // Set the options for the connection
   connection.setKeepAlive(self.keepAlive, self.keepAliveInitialDelay);
   connection.setTimeout(self.connectionTimeout);
   connection.setNoDelay(self.noDelay);
 
-  return connection;
+  // Add handlers for events
+  connection.once('error', err =>
```
mbroadst: for all of these I think its enough to just call `callback(err, null)` . Either we have an error or we have a connection, we don't really care where the error came from at this point, or what type it is.

lib/connection/connection.js
```diff
@@ -528,97 +528,142 @@ function merge(options1, options2) {
...
+    callback({ err: err, type: 'error', family: options.family }, null)
+  );
+  connection.once('timeout', err => callback({ err: err, type: 'timeout' }, null));
+  connection.once('close', err => callback({ err: err, type: 'close' }, null));
+  connection.on('data', err => callback({ err: err, type: 'data' }, null));
```
mbroadst: no need to add a listener for `data`

lib/connection/connection.js
```diff
@@ -528,97 +528,142 @@ function merge(options1, options2) {
...
+  const options = prepareConnectionOptions(self, _options);
+
+  let connection;
+  const connectionHandler = (err, _connection) => {
+    const _errorHandler = errorHandler(self);
+    const _timeoutHandler = timeoutHandler(self);
+    const _closeHandler = closeHandler(self);
+    const _dataHandler = dataHandler(self);
+
+    if (err) {
```
mbroadst: I think if you follow the suggestion above, this whole section just becomes:
```
if (err) {
  if (self.logger.isDebug()) {
    self.logger.debug(
      `connection ${self.id} for [${self.host}:${self.port}] errored out with [${JSON.stringify(error)}]`
    );
  }

  return;
}
```

lib/connection/connection.js
```diff
@@ -528,97 +528,142 @@ function merge(options1, options2) {
...
-    // Emit connect event
-    self.emit('connect', self);
-  });
+function fastFallback(self, _options, callback) {
+  const options = prepareConnectionOptions(self, _options);
+
+  let connection;
+  const connectionHandler = (err, _connection) => {
+    const _errorHandler = errorHandler(self);
+    const _timeoutHandler = timeoutHandler(self);
```
mbroadst: I don't think you need to create these handlers here, they will be created later when you attach them to the returned connection. Here we are _only_ trying to make the connections, focusing on listening to them properly later.

lib/connection/connection.js
shsaskin: This is part of the fast fallback algorithm, by having ipv4 wait 250ms to try to connect, it gives ipv6 preference.

mbroadst: Ah okay! Can you leave a note indicating as much here for future readers?

lib/connection/connection.js
```diff
@@ -528,97 +528,113 @@ function merge(options1, options2) {
...
+    if (_connection) {
+      if (connection) {
+        ['error', 'timeout', 'close'].forEach(event => _connection.removeAllListeners(event));
+        _connection.end();
+        _connection.unref();
+        return;
+      }
+      connection = _connection;
+      return callback(null, connection);
+    } else if (err) {
```
shsaskin: @mbroadst 
Three tests are failing because they expect an error to be emitted here with the line

`if (conn.listeners('error').length > 0) conn.emit('error', new MongoNetworkError(err), conn);`
from the errorHandler function.

However, this functionality should only exist when the family is 4 or the family is passed in (which isn't implemented yet), and when I try to implement this for an ipv4 family, I get the issue of done being called twice in the connection tests.

mbroadst: Right, we still need to emit an error if one is returned. I think this event handler should look something like this:
```
let errors = [];
let connection;
function connectionHandler(err, _connection) => {
  if (connection) {
    // we already have an active connection, clean this one up
    ['error', 'timeout', 'close'].forEach(event => _connection.removeAllListeners(event));
    _connection.end();
    _connection.unref();
    return;
  }

  if (err) {
    if (errors.length > 0) {
      // an error occurred for the second time, we have officially failed
      return callback(new MongoError('failed to connect'), null);  // maybe make a MongoError an attach both errors to it? 
    }

    // otherwise push the error, and wait for subsequent connects
    errors.push(err);
    return;
  }

  // we have a successful connection, set the connection and callback!
  connection = _connection;
  callback(null, connection);
}
```

test/tests/functional/rs_mocks/add_remove_tests.js
```diff
@@ -602,7 +602,7 @@ describe('ReplSet Add Remove (mocks)', function() {
             server.destroy();
             done();
           }, 1000);
-        }, 500);
+        }, 630); // This connection implementation slows down conneciton to nodes, need to fix this
```
mbroadst: do you still need this change?

shsaskin: Yes, this is still necessary, however, I am currently debugging the close event. I am hoping that closing at the proper time will speed up the connections

lib/connection/connection.js
```diff
@@ -528,97 +528,134 @@ function merge(options1, options2) {
...
+    // Set options for ssl
+    if (conn.ca) options.ca = conn.ca;
+    if (conn.crl) options.crl = conn.crl;
+    if (conn.cert) options.cert = conn.cert;
+    if (conn.key) options.key = conn.key;
+    if (conn.passphrase) options.passphrase = conn.passphrase;
+
+    // Override checkServerIdentity behavior
+    if (conn.checkServerIdentity === false) {
+      // Skip the identiy check by retuning undefined as per node documents
```
kvwalker: nit: 'retuning' should be 'returning'

PR #344, opened 2018-08-13
refactor(topologies): rename `getServer` to `selectServer`

mbroadst:
This makes it much easier to drop-in replace the upcoming unified
`Topology` type by making the internal APIs the same.

NODE-1625

### DIFF
--- a/test/tests/functional/mongos_server_selection_tests.js
+++ b/test/tests/functional/mongos_server_selection_tests.js
@@ -1,30 +1,22 @@
 'use strict';
 
-var expect = require('chai').expect,
-  f = require('util').format,
-  fs = require('fs'),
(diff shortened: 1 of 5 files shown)


test/tests/functional/mongos_server_selection_tests.js
```diff
@@ -34,14 +26,14 @@ function convert(mode) {
   return mode.toLowerCase();
 }
 
-function executeEntry(file, path) {
-  console.log('= file :: ' + file);
+function executeEntry(file, path, done) {
+  // console.log('= file :: ' + file);
```
shsaskin: Did you mean to leave this here?

mbroadst: nope! thanks for catching that

PR #345, opened 2018-08-16
fix(EJSON): export the result of optionally requiring EJSON

rweinberger:
Optionally require and export EJSON from core for use in native's APM logging feature.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -48,6 +48,32 @@ try {
   console.warn(err.message);
 }
 
+// Throw an error if an attempt to use EJSON is made when it is not installed
+const noEJSONError = function() {
+  throw new Error(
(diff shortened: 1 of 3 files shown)


lib/utils.js
```diff
@@ -48,6 +48,32 @@ try {
   console.warn(err.message);
 }
 
+// Throw an error if an attempt to use EJSON is made when it is not installed
+const noEJSONError = function() {
+  throw new Error(
+    'APM Logging via URI string requires the `mongodb-extjson` package for parsing event objects, but it was not found. Install the package or disable APM logging and try again.'
```
mbroadst: Let's change this message to be specifically about extended JSON, we will know what to do if someone forgot to optionally install it.

mbroadst: Just that it's not installed, it should be fairly obvious from the backtrace where it's being called from

PR #346, opened 2018-08-18
fix(buffer): use safe-buffer polyfill to maintain compatibility

Fonger:
Hi,

The fix of [NODE-1461](https://jira.mongodb.org/projects/NODE/issues/NODE-1461?filter=allissues) ( commit 7c71e199190a53017b921e88200cdba3889ca23c) break compatibility with node.js < 4.5 and node.js < 5.9 and  and start to affect downstream packages like Automattic/mongoose#6884

This pull request uses `safe-buffer` to polyfill and [it is suggested by official node.js](https://nodejs.org/en/docs/guides/buffer-constructor-deprecation/)

The polyfill won't change any existing behavior and will call native method directly if it is supported.

### DIFF
--- a/lib/auth/scram.js
+++ b/lib/auth/scram.js
@@ -4,7 +4,8 @@ var f = require('util').format,
   crypto = require('crypto'),
   retrieveBSON = require('../connection/utils').retrieveBSON,
   Query = require('../connection/commands').Query,
-  MongoError = require('../error').MongoError;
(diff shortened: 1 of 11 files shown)


mbroadst:
Hi @Fonger, thanks for this. We are indeed deprecating node v4 in the upcoming months, but in the meantime should in fact be using this. Thanks for the contribution!

PR #347, opened 2018-08-27
fix(uri_parser): use admin as default auth.db

kvwalker:
Fixes NODE-1635


mbroadst:
@kvwalker should we bring in the auth spec tests for this?

PR #348, opened 2018-08-28
refactor(auth): refactor auth

daprahamian:
First part of a major refactor. Next parts are:

 + [x] Fully integrating SCRAM into the AuthProvider class (scram is more complicated and requires more refactoring
+ [x] Switch to using MongoCredentials object within all of mongodb-core.


lib/auth/authProvider.js
```diff
@@ -0,0 +1,157 @@
...
+    if (typeof mechanismProperties === 'function') {
+      callback = mechanismProperties;
+      mechanismProperties = undefined;
+    }
+    const credentials = new MongoCredentials({ username, password, source, mechanismProperties });
+    return this._auth(server, connections, credentials, callback);
+  }
+
+  /**
+   * Impl of auth
```
mbroadst: would prefer a better description of what this is doing (5/10)

lib/auth/authProvider.js
```diff
@@ -0,0 +1,157 @@
...
+        } else {
+          if (errorObject == null) {
+            errorObject = new MongoError(`failed to authenticate using ${credentials.mechanism}`);
+          }
+          callback(errorObject, false);
+        }
+      });
+    };
+
+    const _execute = _connection => process.nextTick(() => execute(_connection));
```
mbroadst: would prefer better name than `_execute` (3/10)

lib/auth/authProvider.js
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  /**
+   * Adds credentials to store only if it does not exist
+   * @param {MongoCredentials} credentials credentials to add to store
+   */
+  addCredentials(credentials) {
+    let found = false;
+
+    for (let i = 0; i < this.authStore.length; i++) {
```
mbroadst: recommend `const found = this.authStore.some(cred => cred.equals(credentials))`

lib/auth/authProvider.js
```diff
@@ -0,0 +1,157 @@
...
+        count = count - 1;
+        if (count === 0) {
+          callback(err, null);
+        }
+      });
+    }
+  }
+
+  /**
+   * Remove authStore credentials
```
mbroadst: prefer better description here

lib/auth/authProvider.js
```diff
@@ -0,0 +1,157 @@
+'use strict';
+
+const MongoCredentials = require('./mongoCredentials').MongoCredentials;
+const MongoError = require('../error').MongoError;
+
+/**
+ * Creates a new Authentication mechanism
```
mbroadst: I'm being a bit of a nitpick on documentation, but I think it should be very clear what the design is. Can you provide more color on what an Authentication Mechanism is, and what this base class is meant do.

lib/auth/authProvider.js
```diff
@@ -0,0 +1,157 @@
+'use strict';
```
mbroadst: To be consistent with the rest of the codebase, this file (and mongoCredentials) should be named in `snake-case.js`

lib/auth/authProvider.js
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * This is a result from a authentication strategy
+ *
+ * @callback authResultCallback
+ * @param {error} error An error object. Set to null if no error present
+ * @param {boolean} result The result of the authentication process
+ */
+
+module.exports = { AuthProvider };
```
mbroadst: I don't mind exporting this as an object for a single object, but I think we should perhaps be more consistent about this. Right now I believe we tend to export a direct object if there is only one in a file.

lib/auth/gssapi.js
```diff
@@ -1,114 +1,45 @@
...
- * Authenticate
- * @method
- * @param {{Server}|{ReplSet}|{Mongos}} server Topology the authentication method is being called on
- * @param {[]Connections} connections Connections to authenticate using this authenticator
- * @param {string} db Name of the database
- * @param {string} username Username
- * @param {string} password Password
- * @param {authResultCallback} callback The callback to return the result from the authentication
- * @return {object}
+ * @extends AuthProvider
```
mbroadst: fwiw jsdoc will automatically pick this up when defined using class syntax

lib/auth/gssapi.js
```diff
@@ -1,114 +1,45 @@
...
-          } else if (count === 0) {
-            if (errorObject == null)
-              errorObject = new MongoError(f('failed to authenticate using mongocr'));
-            callback(errorObject, false);
-          }
-        }
-      );
-    };
+class GSSAPI extends AuthProvider {
+  _authenticateSingleConnection(server, connection, credentials, callback) {
```
mbroadst: Can you document this as an `@override`?

lib/auth/gssapi.js
```diff
@@ -1,114 +1,45 @@
...
+      callback
+    );
+  }
 
-    var _execute = function(_connection) {
-      process.nextTick(function() {
-        execute(_connection);
-      });
-    };
+  _auth(server, connections, credentials, callback) {
```
mbroadst: ditto on `@override`

lib/auth/gssapi.js
mbroadst: On a reread, this is actually a method that `AuthProvider` _assumes_ exists, but is not actually defined with a default implementation. I recommend adding a default imply that throws an error

lib/auth/mongoCredentials.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
```
mbroadst: rename => `mongo-credentials.js`

lib/auth/mongoCredentials.js
```diff
@@ -0,0 +1,23 @@
...
+
+class MongoCredentials {
```
mbroadst: class lacks documentation

lib/auth/plain.js
```diff
@@ -1,176 +1,42 @@
...
-};
-
-AuthSession.prototype.equal = function(session) {
-  return (
-    session.db === this.db &&
-    session.username === this.username &&
-    session.password === this.password
-  );
-};
+// TODO: can we get the Binary type from this.bson instead?
```
mbroadst: agree with comment, I think we should actually only ever use `retrieveBSON` in one place, and then use the bson passed down the chain for everything else

daprahamian: This may have to wait until we upgrade mongodb-core to the latest version of BSON. currently `this.bson` is an instance of `BSON`, so it does not have the static members that are defined on `BSON`. Unless we want to do `this.bson.constructor.Binary`?

lib/auth/authProvider.js
daprahamian: Yes, but that is not forward looking. If/when we switch to ES Module syntax, this will make things easier.

lib/auth/mongo_credentials.js
```diff
@@ -0,0 +1,59 @@
...
+    this.mechanism = options.mechanism;
+    this.mechanismProperties = options.mechanismProperties;
+  }
+
+  /**
+   * Determines if two MongoCredentials objects are equivalent
+   * @param {MongoCredentials} other another MongoCredentials object
+   * @returns {boolean} true if the two objects are equal.
+   */
+  equal(other) {
```
mbroadst: nit: prefer `equals`

lib/topologies/server.js
```diff
@@ -924,27 +921,16 @@ Server.prototype.logout = function(dbName, callback) {
...
  * @param {authResultCallback} callback A callback function
  */
-Server.prototype.auth = function(mechanism, db) {
-  var self = this;
-
-  if (mechanism === 'default') {
-    mechanism = getDefaultAuthMechanism(self.ismaster);
+Server.prototype.auth = function(credentials, callback) {
+  if (credentials.mechanism.toLowerCase() === 'default') {
+    credentials.mechanism = getDefaultAuthMechanism(this.ismaster);
```
mbroadst: should credentials be mutated? I'm starting to feel like more of our types should be frozen, and immutable for their lifetime

lib/auth/auth_provider.js
```diff
@@ -0,0 +1,177 @@
...
+  /**
+   * Re authenticate pool
+   * @method
+   * @param {AuthWriteCommand} writeCommand Topology the authentication method is being called on
+   * @param {Connection[]} connections Connections to authenticate using this authenticator
+   * @param {authResultCallback} callback The callback to return the result from the authentication
+   * @return {object}
+   */
+  reauthenticate(writeCommand, connections, callback) {
+    const authStore = this.authStore.slice(0);
```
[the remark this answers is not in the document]
daprahamian: I think it is just making a copy of the authStore.

lib/auth/auth_provider.js
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  /**
+   * Remove credentials that have been previously stored in the auth provider
+   * @method
+   * @param {string} source Name of database we are removing authStore details about
+   * @return {object}
+   */
+  logout(source) {
+    this.authStore = this.authStore.filter(credentials => credentials.source !== source);
```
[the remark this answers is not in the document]
daprahamian: I'm fine leaving it as-is. `native` is used to telling the topology to logout of a specific database, so I don't see a need to force a conversion to a credentials object there (unlike in our `auth` case, where it is constantly passing around a group of fields. But if you think we should switch entirely to credentials, we can

lib/auth/auth_provider.js
```diff
@@ -0,0 +1,177 @@
...
+
+  /**
+   * Authenticate
+   * @method
+   * @param {AuthWriteCommand} writeCommand Topology the authentication method is being called on
+   * @param {Connection[]} connections Connections to authenticate using this authenticator
+   * @param {MongoCredentials} credentials Authentication credentials
+   * @param {authResultCallback} callback The callback to return the result from the authentication
+   */
+  auth(writeCommand, connections, credentials, callback) {
```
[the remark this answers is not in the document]
daprahamian: At this point in time, `authProviders` are shared across an entire topology. So every server (and by extension every pool) inherits an authProvider from the parent replset/mongos. With the way it is currently done, I don't think we can safely do what you are suggesting, but it may be possible in follow-on work.

lib/connection/pool.js
```diff
@@ -215,21 +215,30 @@ function stateTransition(self, newState) {
...
-  if (auth[0] === undefined) return cb(null);
+function authenticate(pool, credentials, connection, cb) {
+  // If there are no credentials, do not authenticated
+  if (!credentials) {
+    return cb(null);
+  }
+
+  // HACK: If default mechanism has not been determined yet, do not
+  // try to authenticate
+  if (credentials.mechanism === 'default') {
```
mbroadst: is this doing what the comment indicates it's doing? This looks like `If the mechanism IS default, do not try to authenticate`

lib/connection/pool.js
```diff
@@ -215,21 +215,30 @@ function stateTransition(self, newState) {
...
     throw new MongoError(f('authMechanism %s not supported', mechanism));
   }
 
   // Get the provider
-  var provider = pool.authProviders[mechanism];
+  const provider = pool.authProviders[mechanism];
 
   // Authenticate using the provided mechanism
-  provider.auth.apply(provider, [write(pool), [connection], db].concat(auth.slice(2)).concat([cb]));
+  provider.auth.call(provider, write(pool), [connection], credentials, cb);
```
mbroadst: Do we really need to `apply` or `call` here? Why can't we just call `provider.auth(...)`?

lib/connection/pool.js
```diff
@@ -687,16 +696,14 @@ Pool.prototype.isDisconnected = function() {
  * Connect pool
  * @method
  */
-Pool.prototype.connect = function() {
+Pool.prototype.connect = function(credentials) {
```
mbroadst: document `credentials` please

lib/connection/pool.js
```diff
@@ -687,16 +696,14 @@ Pool.prototype.isDisconnected = function() {
...
   if (this.state !== DISCONNECTED) {
     throw new MongoError('connection in unlawful state ' + this.state);
   }
 
   var self = this;
   // Transition to connecting state
   stateTransition(this, CONNECTING);
-  // Create an array of the arguments
```
mbroadst: always _love_ removing code like this :+1:

lib/connection/pool.js
```diff
@@ -783,10 +790,9 @@ Pool.prototype.connect = function() {
  * @param {...object} param Parameters for the specific mechanism
  * @param {authResultCallback} callback A callback function
  */
-Pool.prototype.auth = function(mechanism) {
+Pool.prototype.auth = function(credentials, callback) {
```
mbroadst: Love the consistency of the `credentials` refactor! Please document the new parameter though

lib/connection/pool.js
```diff
@@ -783,10 +790,9 @@ Pool.prototype.connect = function() {
...
   var self = this;
-  var args = Array.prototype.slice.call(arguments, 0);
-  var callback = args.pop();
+  const mechanism = credentials ? credentials.mechanism : '';
```
mbroadst: seems to me we should rather bail if no `credentials` were provided? Isn't there always a mechanism, i.e. `default` if none was provided

lib/topologies/mongos.js
```diff
@@ -661,17 +662,11 @@ function applyAuthenticationContexts(self, server, callback) {
...
-    // Get the first auth context
-    var authContext = authContexts.shift();
-    // Copy the params
-    var customAuthContext = authContext.slice(0);
-    // Push our callback handler
-    customAuthContext.push(function(/* err */) {
-      applyAuth(authContexts, server, callback);
-    });
+    // Get credentials
+    const credentials = authContexts.shift();
```
mbroadst: should `authContexts` be renamed? This looks weird that we shift a credential off the auth context

lib/topologies/mongos.js
```diff
@@ -1133,12 +1128,9 @@ Mongos.prototype.cursor = function(ns, cmd, options) {
  * @param {...object} param Parameters for the specific mechanism
  * @param {authResultCallback} callback A callback function
  */
-Mongos.prototype.auth = function(mechanism, db) {
-  var allArgs = Array.prototype.slice.call(arguments, 0).slice(0);
-  var self = this;
-  var args = Array.prototype.slice.call(arguments, 2);
-  var callback = args.pop();
-  var currentContextIndex = 0;
+Mongos.prototype.auth = function(credentials, callback) {
```
mbroadst: documentation no longer lines up with the method signature

mbroadst: also should we be doing validation of the parameters here?

lib/topologies/mongos.js
```diff
@@ -1152,71 +1144,72 @@ Mongos.prototype.auth = function(mechanism, db) {
...
+
+  // Get total count
+  let count = servers.length;
+
   // No servers return
-  if (servers.length === 0) {
+  if (count === 0) {
     this.authenticating = false;
-    callback(null, true);
+    return callback(null, true);
```
mbroadst: wow looks like that was a long outstanding bug, are you sure we haven't come to rely on this behavior?

test/tests/functional/basic_single_server_auth_tests.js
```diff
@@ -7,29 +7,37 @@ var expect = require('chai').expect,
...
         pwd: password,
-        roles: [{ role: 'root', db: 'admin' }],
+        roles: [{ role: 'root', db: source }],
         digestPassword: true
       };
-      const dropUserCommand = { dropUser: user };
-      const auth = [method, 'admin', user, password];
+      const dropUserCommand = { dropUser: username };
+
+      const credentials = new MongoCredentials({
```
mbroadst: question: these can all be expressed by connection string, shouldn't we use that? Follow up question: should we provide a means of extracting `MongoCredentials` from a connection string, and/or simply have the connection string parser output credentials for the `auth` object?

mbroadst: Final thought: this might be the case for not using the key `credentials`. The URI parser, by spec, parses into an `auth` key so we'd have to do translation there. Food for thought.

lib/connection/pool.js
daprahamian: doh 🤦‍♂️ . You're right.

lib/connection/pool.js
daprahamian: The explanation for this is a little complicated:

+ If you do not provide a mechanism, or if you say "default" we need do to mechanism negotiation. As of MongoDB 4.0, this requires an `ismaster` to get `saslSupportedMechs`
+ For mechanism negotiation to work, we need to have credentials available at the time of topology connect, which happens before the actual `auth` call. This way, we can add the `saslSupportedMechs` field to the first outgoing `ismaster`
+ At the same time, there is legacy support for authenticating on connect, where you pass credentials in to the initial connect call, and immediately auth. This is a problem, since the credentials have not been mechanism negotiated yet, so they will still say `default`.

TBH I'm not happy about the workflow here, and I'd like to brainstorm some solutions for it. But this is currently mimicking the original behavior.

lib/topologies/mongos.js
daprahamian: I really really hope not

test/tests/functional/basic_single_server_auth_tests.js
daprahamian: I've added helper functions in [the native PR](https://github.com/mongodb/node-mongodb-native/pull/1826) to do this. Rather than parse auth credentials out of the connection string, I have a function that parses the output of the url parser into credentials. My thinking is that, since native supports any combination of url and natively passed options, it is better to wait until both url options and natively passed options are merged.

lib/connection/pool.js
mbroadst: ok, two things: can you update the note to better reflect what's going on here (just so I don't make the same mistake when reading through again), and secondly can you create a ticket to track refactoring this process, and add it to the Auth Refactor epic

lib/topologies/server.js
daprahamian: Part of the benefit of mutation is that you can resolve it only once. Otherwise, someone can hold on to stale credentials, and they will need to re-resolve the negotiation each time the credentials are presented.

lib/connection/pool.js
daprahamian: Issue is [here](https://jira.mongodb.org/browse/NODE-1674)

lib/connection/pool.js
daprahamian: We do not validate this far down. TBH anything that isn't a valid authprovider should have errored either in native or in the server by the time it gets here.

lib/topologies/mongos.js
daprahamian: I don't think we should? Why would you `auth` with no credentials?

mbroadst: 🤷‍♂️ , I clearly have PTSD from our general lack of options validation 😓

test/tests/functional/basic_single_server_auth_tests.js
mbroadst: SGTM

lib/auth/auth_provider.js
```diff
@@ -0,0 +1,177 @@
...
+ * mechanism.
+ * @class
+ */
+class AuthProvider {
+  constructor(bson) {
+    this.bson = bson;
+    this.authStore = [];
+  }
+
+  _buildCommand(namespace, cmd) {
```
mbroadst: lacks docs, I do expect that this abstracts creating a Query so that's cool (for e.g. OP_MSG in the future)

lib/auth/auth_provider.js
```diff
@@ -0,0 +1,177 @@
...
+   * @param {Connection[]} connections Connections to authenticate using this authenticator
+   * @param {MongoCredentials} credentials Authentication credentials
+   * @param {authResultCallback} callback The callback to return the result from the authentication
+   */
+  auth(sendAuthCommand, connections, credentials, callback) {
+    return this._auth(sendAuthCommand, connections, credentials, callback);
+  }
+
+  /**
+   * Implementation of auth. Private method. expect this to go away once refactor is complete
```
mbroadst: Is this still going away as part of the refactor?

lib/auth/auth_provider.js
```diff
@@ -0,0 +1,177 @@
...
+
+    if (!found) {
+      this.authStore.push(credentials);
+    }
+  }
+
+  /**
+   * Re authenticate pool
+   * @method
+   * @param {SendAuthCommand} sendAuthCommand Topology the authentication method is being called on
```
mbroadst: incorrect documentation

lib/auth/mongo_credentials.js
```diff
@@ -0,0 +1,81 @@
...
+    );
+  }
+
+  /**
+   * If the authentication mechanism is set to "default", resolves the authMechanism
+   * based on the server version and server supported sasl mechanisms.
+   *
+   * @param {Object} [ismaster] An ismaster response from the server
+   */
+  resolveAuthMechanism(ismaster) {
```
mbroadst: love it!

lib/auth/mongocr.js
```diff
@@ -1,214 +1,63 @@
...
-                callback(null, true);
-              } else if (count === 0) {
-                if (errorObject == null)
-                  errorObject = new MongoError(f('failed to authenticate using mongocr'));
-                callback(errorObject, false);
-              }
-            }
-          );
-        }
+      const authenticateCommand = new Query(
```
mbroadst: why isn't this using `buildCommand`?

lib/auth/mongocr.js
```diff
@@ -1,214 +1,63 @@
...
+   * Implementation of authentication for a single connection
+   * @override
+   */
+  _authenticateSingleConnection(writeCommand, connection, credentials, callback) {
+    const username = credentials.username;
+    const password = credentials.password;
+    const source = credentials.source;
+
+    // Write the commmand on the connection
+    const getNonceCommand = new Query(
```
mbroadst: ditton on `buildCommand`

lib/auth/mongocr.js
```diff
@@ -1,214 +1,57 @@
...
+    const source = credentials.source;
+
+    // Write the commmand on the connection
+    const getNonceCommand = this._buildCommand(`${source}.$cmd`, { getnonce: 1 });
+
+    writeCommand(connection, getNonceCommand, (err, r) => {
+      let nonce = null;
+      let key = null;
+
+      // Adjust the number of connections left
```
kvwalker: Where is the number of connections left being adjusted?

lib/connection/pool.js
```diff
@@ -215,21 +215,32 @@ function stateTransition(self, newState) {
   }
 }
 
-function authenticate(pool, auth, connection, cb) {
-  if (auth[0] === undefined) return cb(null);
+function authenticate(pool, credentials, connection, cb) {
+  // If there are no credentials, do not authenticated
```
kvwalker: nit: 'authenticated' should be 'authenticate'

PR #358, opened 2018-09-17
chore(evg): drop zSeries for all but rhel, reintroduce rhel72

mbroadst:
NODE-1679

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -386,11 +386,6 @@ axes:
   - id: os-fully-featured
     display_name: OS
     values:
-      # - id: linux-64-amzn-test
-      #   display_name: "Amazon Linux (Enterprise)"
-      #   run_on: linux-64-amzn-test
-      #   batchtime: 10080  # 7 days
-
       - id: ubuntu-14.04
         display_name: "Ubuntu 14.04"
         run_on: ubuntu1404-test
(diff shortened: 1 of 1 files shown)


.evergreen/config.yml
```diff
@@ -456,14 +446,9 @@ axes:
...
       # NOTE: reenable when these are actually running 7.2, or we release a 7.4 rpm
-      # - id: rhel72-zseries-test
-      #   display_name: "RHEL 7.2 (zSeries)"
-      #   run_on: rhel72-zseries-test
-      #   batchtime: 10080  # 7 days
-
-      - id: suse12-zseries-test
-        display_name: "SUSE 12 (zSeries)"
-        run_on: suse12-zseries-test
+      - id: rhel72-zseries-test
```
kvwalker: Should this be `rhel67-zseries-test`, which is what the [drivers ticket ](https://jira.mongodb.org/browse/DRIVERS-469) says?

mbroadst: yep! nice catch thanks

mbroadst: hm looks like we can't actually build over there due to gcc version, let's let this PR sit until someone on dev prod can update on that

mbroadst: @kvwalker it was determined we should stick with rhel72, reverted

PR #359, opened 2018-09-21
fix(uri-parser): validate connection string options

mbroadst:
This primarily introduces a translation layer for the new URI parser to convert known options back to camelCase from their normalized lowercase form (expected by the connection string specification). A number of other related fixes are included. 

Fixes NODE-1681


lib/uri_parser.js
```diff
@@ -161,13 +162,146 @@ function parseQueryStringItemValue(value) {
   return value;
 }
 
+// Options that are known boolean types
+const BOOLEAN_OPTIONS = new Set([
+  'slaveok',
+  'slave_ok',
+  'sslvalidate',
+  'fsync',
+  'safe',
```
daprahamian: `safe` is definitely not supported anymore, and I think the same is true for `fsync`. Also, I am surprised that we still support master/slave terminology.

lib/uri_parser.js
```diff
@@ -161,13 +162,146 @@ function parseQueryStringItemValue(value) {
...
+  'MONGODB-CR',
+  'DEFAULT',
+  'SCRAM-SHA-1',
+  'SCRAM-SHA-256',
+  'PLAIN'
+]);
+
+// Lookup table used to translate normalized (lower-cased) forms of connection string
+// options to their expected camelCase version
+const CASE_TRANSLATION = {
```
daprahamian: I feel like we do this sort of thing all over the place. For the future, can we create a single location for all of our options where we specify metadata about them?

PR #361, opened 2018-09-23
refactor(auth): remove `auth` api endpoints

mbroadst:
This completely removes the ability to authenticate through the "public" api of the topologies in core. This was a holdover from a time when we allowed authentication against multiple databases for a single topology, but have not technically allowed this since the 3.x bump.

Authentication is now completely handled by passing a `MongoCredentials` object in the options of a topology, and those credentials are individually applied to `Server` instances at the time they initially connect. This handles all cases: when using a single server instance, as well as the container topologies `Mongos` and `ReplSet`.

This was a bit more of an intrusive change than anticipated, and required refactoring the topologies we hope to completely deprecate very soon, but should make implementation authentication in the new SDAM layer a cakewalk.

Fixes NODE-1674


lib/topologies/replset.js
```diff
@@ -292,7 +292,7 @@ function connectNewServers(self, servers, callback) {
 
   // Handle events
   var _handleEvent = function(self, event) {
-    return function(err) {
+    return function(_, err) {
```
daprahamian: why are we adding this spacer here?

mbroadst: because I broke "BC" by adding the error as the first parameter emitted by the event rather than the topology. This keeps it from breaking anyone who might have been using core before, and requires less changes to the tests

PR #363, opened 2018-09-24
refactor(uri-parser): support authentication expectations

mbroadst:
Adding support for the expectations put forth in the authentication spec, as well as the accompanied tests.

Fixes NODE-1689


lib/uri_parser.js
```diff
@@ -296,6 +296,76 @@ function applyConnectionStringOption(obj, key, value, options) {
...
+      parsed.auth = Object.assign({}, parsed.auth, { db: '$external' });
+    }
+
+    if (authMechanism === 'MONGODB-X509') {
+      if (parsed.auth && parsed.auth.password != null) {
+        throw new MongoParseError(`Password not allowed for mechanism \`${authMechanism}\``);
+      }
+
+      if (authSource != null && authSource !== '$external') {
+        throw new MongoParseError(`Invalid auth source \`${authMechanism}\` specified`);
```
kvwalker: Same note as above.

lib/uri_parser.js
```diff
@@ -296,6 +296,76 @@ function applyConnectionStringOption(obj, key, value, options) {
...
+    if (
+      USERNAME_REQUIRED_MECHANISMS.has(authMechanism) &&
+      (!parsed.auth || parsed.auth.username == null)
+    ) {
+      throw new MongoParseError(`Username required for mechanism \`${authMechanism}\``);
+    }
+
+    if (authMechanism === 'GSSAPI') {
+      if (authSource != null && authSource !== '$external') {
+        throw new MongoParseError(`Invalid auth source \`${authMechanism}\` specified`);
```
kvwalker: We could make this error message more clear, like `Invalid authSource ${authSource} for mechanism ${authMechanism} specified.`

lib/uri_parser.js
```diff
@@ -296,6 +296,76 @@ function applyConnectionStringOption(obj, key, value, options) {
...
+  if (authMechanism != null) {
+    if (
+      USERNAME_REQUIRED_MECHANISMS.has(authMechanism) &&
+      (!parsed.auth || parsed.auth.username == null)
+    ) {
+      throw new MongoParseError(`Username required for mechanism \`${authMechanism}\``);
+    }
+
+    if (authMechanism === 'GSSAPI') {
+      if (authSource != null && authSource !== '$external') {
```
daprahamian: Does this rule also apply to `X509`?

daprahamian: NVM you did it below :)

lib/uri_parser.js
```diff
@@ -296,6 +296,76 @@ function applyConnectionStringOption(obj, key, value, options) {
...
+  }
+
+  const options = parsed.options;
+  const authSource = options.authsource || options.authSource;
+  if (authSource != null) {
+    parsed.auth = Object.assign({}, parsed.auth, { db: authSource });
+  }
+
+  const authMechanism = options.authmechanism || options.authMechanism;
+  if (authMechanism != null) {
```
daprahamian: Can we break up some of the sub-`if`s into sub-functions? This `if` seems really big

test/tests/unit/mongodb_srv_tests.js
```diff
@@ -0,0 +1,63 @@
+'use strict';
```
daprahamian: Double checking: this is just renaming `mongodb_srv_spec_tests.js`?

mbroadst: it does that, but also split the non-spec test involved into its own `describe` separate from the spec tests, and fixes that test not to use a SRV host that returns an `authSource`

lib/uri_parser.js
mbroadst: This is what I was referring to yesterday in our meeting. I think all of these, in a more elegantly designed system, would live in some sort of factory method on each of the sasl modules. As it stands, its a little long but its a literal parse and I'd prefer to actually reduce the function calls unless its bringing a great increase in readability. Do you feel strongly about this?

test/tests/unit/mongodb_srv_tests.js
daprahamian: nah, if we moved it to the authproviders later that would be fine. Can we file a ticket for it?

lib/uri_parser.js
daprahamian: Put this comment on the wrong thread: nah, if we moved it to the authproviders later that would be fine. Can we file a ticket for it?

mbroadst: Done. [NODE-1691](https://jira.mongodb.org/browse/NODE-1691)

PR #364, opened 2018-09-26
refactor(server): introduce authentication to server class

mbroadst:
This enables authentication for the new SDAM layer with a relatively minor changeset!

NODE-1620


mbroadst:
@daprahamian there are a few tests in `pool_tests.js`, but generally yes I'm depending on native to test this. It's my hope that in the future (especially once `Topology` just supports a connection string) that we move most of those tests from native into core, ensuring that we are covering the right code in the right project.

PR #365, opened 2018-10-01
fix(mongos): solve memory leak on server reconnection

itayalfia:
**The closure of the `serverDescriptionChanged` event handler references the old _server instance, preventing old disconnected server instances from being cleaned by GC.
Has noticeable effect on memory usage after a large amount of disconnections.

Solution: bind the emit function to its parameters without declaring a function which would inherit the closure.**

Can be tested by changing the test in mongodb-core\test\tests\unit\mongos\reconnect_tests.js to take heapdumps.

Attached 2 heapdumps, with and without the fix. When opened in chrome dev-tools, the version without the fix contains a larger amount of Server instances, retained by serverDescriptionChanged event.

[heapdumps.zip](https://github.com/mongodb-js/mongodb-core/files/2435540/heapdumps.zip)
example-mem

### DIFF
--- a/lib/topologies/mongos.js
+++ b/lib/topologies/mongos.js
@@ -617,9 +617,7 @@ function reconnectProxies(self, proxies, callback) {
       removeProxyFrom(self.disconnectedProxies, _server);
 
       // Relay the server description change
-      server.on('serverDescriptionChanged', function(event) {
-        self.emit('serverDescriptionChanged', event);
-      });
+      server.on('serverDescriptionChanged', self.emit.bind(self, 'serverDescriptionChanged'));
 
       // Emit opening server event
       self.emit('serverOpening', {


itayalfia:
@mbroadst Hi, can you please give me feedback for my PR? Is everything ok with my fix?

mbroadst:
@itayalfia we're just missing a test case here, I'm working on putting one together locally. Will get back to you shortly, thanks!

mbroadst:
@itayalfia I see, this is just a matter of a left-over event listener really. We don't need a test, but I think the preferred solution is to add:
```
_server.removeAllListeners('serverDescriptionChanged');
```
on L117, after ` _server.destroy();` Could you make that change? We'll merge it in as soon as that's in, thanks!

PR #367, opened 2018-10-28
fix(uri_parser) bypass number coercion for replicaSet val in connection string

lineus:
re: Automattic/mongoose/issues/7167 ( replicaSet details and native code sample there ).

If a setName starts with a number, the new url parser will coerce the replicaSet value passed into the connection string to a number e.g. `replicaSet=456abc` becomes `456`.  Setting the value in the connection options works as expected. This change adds replicaset to the set of known strings.


mbroadst:
Thanks @lineus, we should have a new release this week including this change.

PR #368, opened 2018-10-30
fix(utils): remove warning for missing kerberos module

kvwalker:
Fixes [NODE-1626](https://jira.mongodb.org/browse/NODE-1626)

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -45,8 +45,8 @@ try {
     MongoAuthProcess = kerberos.processes.MongoAuthProcess;
   }
 } catch (err) {
-  if (err.code !== 'MODULE_NOT_FOUND') {
-      throw err;
+  if (err.code !== "MODULE_NOT_FOUND") {
+    throw err;
   }
 }
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -45,7 +45,7 @@ try {
     MongoAuthProcess = kerberos.processes.MongoAuthProcess;
   }
 } catch (err) {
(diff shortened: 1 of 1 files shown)


daprahamian:
Doesn't this leave us back in the situation where a kerberos require error can be swallowed and confuse people?

kvwalker:
@daprahamian, I thought there was a reason this was added in the first place...do you remember what the ticket was for that problem? In this case, it seems preferable to keep the warning the way it currently is.

daprahamian:
@kvwalker I'm not sure. The [commit](https://github.com/mongodb-js/mongodb-core/commit/b553ee17e7377b4943af29ce0847598a246ed6f2#diff-50e3aa130a4f97a42ee2cf111c7b1d9d) is not really helpful. But I think a better approach might be to store the Error, and then if anyone tries to auth with Kerberos throw the error?

mbroadst:
@daprahamian I think what we should do here is ignore any error related to the requiring of the module, and rethrow other errors. Users should not have to pay the price here for what we are already calling an optional requirement. The problem with the code as its written is that it will _also_ swallow any errors caused during the import (this bit me yesterday, a syntax error resulted in a message that kerberos was missing). Are there other classes or error you are concerned about being confusing?

kvwalker:
@daprahamian, we should be able to check if `error.code === 'MODULE_NOT_FOUND'`, according to https://nodejs.org/api/modules.html#modules_file_modules.

daprahamian:
Okay, then lets do that and test it.

kvwalker:
Closing this for Matt's refactor.

PR #369, opened 2018-10-30
fix(sspi): correct auth process for SSPI

mbroadst:
The SSPI auth process should have been the same as GSSAPI, but
it instead initialized and transitioned kerberos _after_ an
initial connect to the client. Starting with MongoDB 4.0, this
simply will not work anymore, and needs to be corrected.

NODE-1479


mbroadst:
Yep, please make a jira ticket for that. The customer complaint was that linux worked perfectly, and sspi did not, so I didn't want to touch the working path.

PR #372, opened 2018-11-05
refactor(kerberos): throw runtime error for missing kerberos module

mbroadst:
supersedes #368

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -45,8 +45,8 @@ try {
     MongoAuthProcess = kerberos.processes.MongoAuthProcess;
   }
 } catch (err) {
-  if (err.code !== 'MODULE_NOT_FOUND') {
-      throw err;
+  if (err.code !== "MODULE_NOT_FOUND") {
+    throw err;
   }
 }
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -45,7 +45,7 @@ try {
     MongoAuthProcess = kerberos.processes.MongoAuthProcess;
   }
 } catch (err) {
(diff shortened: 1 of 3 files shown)


lib/utils.js
```diff
@@ -34,27 +34,27 @@ function relayEvents(listener, emitter, events) {
...
 
-try {
-  const kerberos = requireOptional('kerberos');
-  if (kerberos) {
-    Kerberos = kerberos.Kerberos;
-    MongoAuthProcess = kerberos.processes.MongoAuthProcess;
+  try {
+    kerberos = requireOptional('kerberos');
+  } catch (err) {
+    if (err.code !== 'MODULE_NOT_FOUND') {
```
daprahamian: If we are now deferring requiring `kerberos` until we actually auth with it, then we shouldn't swallow `MODULE_NOT_FOUND`. If the module isn't found, auth should error.

lib/utils.js
```diff
@@ -34,27 +34,27 @@ function relayEvents(listener, emitter, events) {
...
+      throw new Error('The `kerberos` module was not found. Please install it any try again.');
```
kvwalker: Is this throwing an error for `kerberos not found` for errors not related to `MODULE_NOT_FOUND`?

lib/utils.js
mbroadst: We're swallowing it to control the text of the returned error to be something more user friendly than "module not found"

lib/utils.js
mbroadst: oops, good catch!

PR #383, opened 2018-12-09
refactor(cursor): merge and simplify `nextFunction`

mbroadst:
At first glance this just looks like its merging two functions into
a single, unweildy mega-function... which is sort of what it does.
But! This does a few things for us:

  - reduces an implicit required recursive call for every cursor
    by combining the initialization code with a call to the first
    find command

  - places the preparation of the find command (from WireProtocol)
    with the actual execution (`pool.write`), which allows us to
    better imagine moving those steps into the wireprotocol in
    lock-step. This is good because it takes us one step closer
    to actually having a consistent wire protocol api, and ceasing
    rogue command construction in the codebase

NODE-1785

### DIFF
--- a/lib/cursor.js
+++ b/lib/cursor.js
@@ -209,124 +209,6 @@ var handleCallback = function(callback, err, result) {
 };
 
 // Internal methods
-Cursor.prototype._find = function(callback) {
-  var self = this;
-
-  if (self.logger.isDebug()) {
-    self.logger.debug(
-      f(
-        'issue initial query [%s] with flags [%s]',
-        JSON.stringify(self.cmd),
-        JSON.stringify(self.query)
-      )
-    );
-  }
-
-  var queryCallback = function(err, r) {
-    if (err) return callback(err);
-
-    // Get the raw message
-    var result = r.message;
-
-    // Query failure bit set
-    if (result.queryFailure) {
-      return callback(new MongoError(result.documents[0]), null);
-    }
-
-    // Check if we have a command cursor
-    if (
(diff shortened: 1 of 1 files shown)


lib/cursor.js
```diff
@@ -749,6 +596,41 @@ var nextFunction = function(self, callback) {
   }
 };
 
+function buildFindCommandQueryOptions(cursor) {
+  const queryOptions = {};
+
+  // If we have a raw query decorate the function
+  if (cursor.options.raw || cursor.cmd.raw) {
+    queryOptions.raw = cursor.options.raw || cursor.cmd.raw;
```
kvwalker: I know this was here before, but what happens if `cursor.options.raw` and `cursor.cmd.raw` are set to conflicting values? Should we always be setting `queryOptions.raw` to `true` if at least one of those is `true`?

lib/cursor.js
```diff
@@ -778,49 +660,135 @@ function initializeCursor(cursor, callback) {
...
+      if (
+        Array.isArray(result.documents) &&
+        result.documents.length === 1 &&
+        (!cursor.cmd.find || (cursor.cmd.find && cursor.cmd.virtual === false)) &&
+        (typeof result.documents[0].cursor !== 'string' ||
+          result.documents[0]['$err'] ||
+          result.documents[0]['errmsg'] ||
+          Array.isArray(result.documents[0].result))
+      ) {
+        // We have a an error document return the error
```
kvwalker: delete 'a'

PR #384, opened 2018-12-12
refactor(wire-protocol): modernize and clean wire protocol

mbroadst:
This is a mostly cosmetic change to make the two wire protocol classes easier to reason about:
- refactor to use a common `raw` property on the cursor state
- utilizing a common function for determining common query options
- converting the wire protocols to classes to make them easier to read and determine common code

### DIFF
--- a/lib/cursor.js
+++ b/lib/cursor.js
@@ -7,9 +7,10 @@ const MongoNetworkError = require('./error').MongoNetworkError;
 const mongoErrorContextSymbol = require('./error').mongoErrorContextSymbol;
 const f = require('util').format;
 const collationNotSupported = require('./utils').collationNotSupported;
+const applyCommonQueryOptions = require('./wireprotocol/shared').applyCommonQueryOptions;
 
-var BSON = retrieveBSON(),
-  Long = BSON.Long;
+const BSON = retrieveBSON();
+const Long = BSON.Long;
 
 /**
  * This is a cursor results callback
@@ -105,7 +106,8 @@ var Cursor = function(bson, ns, cmd, options, topology, topologyOptions) {
     batchSize: options.batchSize || cmd.batchSize || 1000,
     currentLimit: 0,
     // Result field name if not a cursor (contains the array of results)
-    transforms: options.transforms
+    transforms: options.transforms,
+    raw: options.raw || (typeof cmd === 'object' && cmd.raw)
   };
 
   if (typeof options.session === 'object') {
@@ -596,41 +598,6 @@ var nextFunction = function(self, callback) {
   }
 };
 
-function buildFindCommandQueryOptions(cursor) {
-  const queryOptions = {};
-
-  // If we have a raw query decorate the function
-  if (cursor.options.raw || cursor.cmd.raw) {
-    queryOptions.raw = cursor.options.raw || cursor.cmd.raw;
-  }
-
-  // Do we have documentsReturnedIn set on the query
-  if (typeof cursor.query.documentsReturnedIn === 'string') {
-    queryOptions.documentsReturnedIn = cursor.query.documentsReturnedIn;
-  }
-
-  // Add promote Long value if defined
-  if (typeof cursor.cursorState.promoteLongs === 'boolean') {
-    queryOptions.promoteLongs = cursor.cursorState.promoteLongs;
-  }
-
-  // Add promote values if defined
-  if (typeof cursor.cursorState.promoteValues === 'boolean') {
-    queryOptions.promoteValues = cursor.cursorState.promoteValues;
-  }
-
-  // Add promote values if defined
-  if (typeof cursor.cursorState.promoteBuffers === 'boolean') {
-    queryOptions.promoteBuffers = cursor.cursorState.promoteBuffers;
-  }
-
-  if (typeof cursor.cursorState.session === 'object') {
-    queryOptions.session = cursor.cursorState.session;
-  }
-
-  return queryOptions;
-}
-
 function initializeCursor(cursor, callback) {
   // Topology is not connected, save the call in the provided store to be
   // Executed at some point when the handler deems it's reconnected
@@ -722,7 +689,12 @@ function initializeCursor(cursor, callback) {
       );
     }
 
-    const queryOptions = buildFindCommandQueryOptions(cursor);
+    const queryOptions = applyCommonQueryOptions({}, cursor.cursorState);
+    // Do we have documentsReturnedIn set on the query
+    if (typeof cursor.query.documentsReturnedIn === 'string') {
+      queryOptions.documentsReturnedIn = cursor.query.documentsReturnedIn;
+    }
+
     cursor.server.s.pool.write(cursor.query, queryOptions, (err, r) => {
       if (err) return callback(err);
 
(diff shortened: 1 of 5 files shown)


mbroadst:
@daprahamian if you could start your review here it would be better for subsequent reviews, they are all based on this one

lib/cursor.js
```diff
@@ -105,7 +106,8 @@ var Cursor = function(bson, ns, cmd, options, topology, topologyOptions) {
     batchSize: options.batchSize || cmd.batchSize || 1000,
     currentLimit: 0,
     // Result field name if not a cursor (contains the array of results)
-    transforms: options.transforms
+    transforms: options.transforms,
+    raw: options.raw || (typeof cmd === 'object' && cmd.raw)
```
daprahamian:
```js
raw: options.raw || (cmd && cmd.raw)
```

lib/wireprotocol/2_6_support.js
```diff
@@ -1,207 +1,140 @@
...
+  }
+
+  getMore(bson, ns, cursorState, batchSize, connection, options, callback) {
+    const getMore = new GetMore(bson, ns, cursorState.cursorId, { numberToReturn: batchSize });
+    function queryCallback(err, result) {
+      if (err) return callback(err);
+      const r = result.message;
+
+      // If we have a timed out query or a cursor that was killed
+      if ((r.responseFlags & (1 << 0)) !== 0) {
```
daprahamian: I'm guessing that this is left over from before, but why the `1 << 0` instead of just `1`?

lib/wireprotocol/2_6_support.js
```diff
@@ -275,94 +198,54 @@ var setupClassicFind = function(bson, ns, cmd, cursorState, topology, options) {
...
-  }
-
-  if (typeof cmd.awaitData === 'boolean') {
-    query.awaitData = cmd.awaitData;
-  }
-
-  if (typeof cmd.partial === 'boolean') {
-    query.partial = cmd.partial;
-  }
+  if (typeof cmd.tailable === 'boolean') query.tailable = cmd.tailable;
```
daprahamian: I'd like to keep the braces if possible.

lib/wireprotocol/2_6_support.js
```diff
@@ -275,94 +198,54 @@ var setupClassicFind = function(bson, ns, cmd, cursorState, topology, options) {
...
-//
-// Set up a command cursor
-var setupCommand = function(bson, ns, cmd, cursorState, topology, options) {
-  // Set empty options object
+function setupCommand(bson, ns, cmd, cursorState, topology, options) {
   options = options || {};
-  // Get the readPreference
-  var readPreference = getReadPreference(cmd, options);
+  const readPreference = getReadPreference(cmd, options);
+  const parts = ns.split(/\./);
```
daprahamian: probably exists from before, but why is this a regex and not just `'.'`?

lib/wireprotocol/shared.js
```diff
@@ -47,9 +47,22 @@ var parseHeader = function(message) {
   };
 };
 
+function applyCommonQueryOptions(queryOptions, cursorState) {
+  if (cursorState.raw) queryOptions.raw = cursorState.raw;
+  if (typeof cursorState.promoteLongs === 'boolean')
```
daprahamian: These are even going on to the next line here, so let's do `{}`

lib/wireprotocol/2_6_support.js
mbroadst: sure

mbroadst: out of curiosity, where do you draw the line? I notice you're not asking to introduce braces above for e.g. all of the one-liners in `setupClassicFind` like:
```
if (cmd.sort) findCmd['$orderby'] = cmd.sort;
```

lib/wireprotocol/2_6_support.js
mbroadst: believe me this is all changing by the 5th PR on top of this one 😄 I refactored this all into `databaseNamespace` and `collectionNamespace`, all of the splitting and joining is hidden away and done without regular expressions

lib/wireprotocol/shared.js
mbroadst: shall we make a rule that braces are only required if you are breaking the line?

kvwalker: I like that rule.

lib/wireprotocol/2_6_support.js
daprahamian: nice :)

lib/wireprotocol/2_6_support.js
mbroadst: Definitely left from before, but I wonder if we're missing more here. The [OP_REPLY](https://docs.mongodb.com/manual/reference/mongodb-wire-protocol/#op-reply) documentation indicates the following for `responseFlags`:
```
0 corresponds to CursorNotFound. Is set when getMore is called but the cursor id is not valid at the server. Returned with zero results.
1 corresponds to QueryFailure. Is set when query failed. Results consist of one document containing an “$err” field describing the failure.
2 corresponds to ShardConfigStale. Drivers should ignore this. Only mongos will ever see this set, in which case, it needs to update config from the server.
3 corresponds to AwaitCapable. Is set when the server supports the AwaitData Query option. If it doesn’t, a client should sleep a little between getMore’s of a Tailable cursor. Mongod version 1.6 supports AwaitData and thus always sets AwaitCapable.
```

but we have them defined in [commands.js](https://github.com/mongodb-js/mongodb-core/blob/master/lib/connection/commands.js#L25-L28) as:
```
var CURSOR_NOT_FOUND = 0;
var QUERY_FAILURE = 2;
var SHARD_CONFIG_STALE = 4;
var AWAIT_CAPABLE = 8;
```

Looks like we're checking for an invalid bit here

lib/wireprotocol/2_6_support.js
mbroadst: Well, taking a look at your [OP_MSG implementation](https://github.com/mongodb-js/mongodb-core/pull/276/files) I see that you use this form, as well as with braces, as well as looping over an array of properties for assigning. 

I think we can all be a little inconsistent, so maybe better if we all agree on a rule of thumb?

lib/wireprotocol/2_6_support.js
mbroadst: Okay, after much discussion and investigation I've identified a few points of interest:
  - it doesn't matter that we're defining `CURSOR_NOT_FOUND` as `0` in `commands.js`, because we only ever check the response flag in the wire protocol files, and there we check the correct value (which is `1`). 

  - half of those flags don't matter anymore, as they were used for feature detection in early versions of the server. 

Nonetheless, I will export the values and use them by name with a corrected value.

lib/wireprotocol/3_2_support.js
```diff
@@ -46,23 +46,23 @@ class WireProtocol {
         return callback(err);
       }
 
-      const r = result.message;
+      const response = result.message;
 
       // If we have a timed out query, or a cursor that was killed
-      if ((r.responseFlags & (1 << 0)) !== 0) {
+      if (response.cursorNotFounds) {
```
kvwalker: Should be `cursorNotFound`

lib/wireprotocol/shared.js
mbroadst: OK @daprahamian, I've corrected the cases where if statements were multi-line. Let's not hang up this review on style choices, but rather punt to the team meeting to discuss potentially adding an eslint/prettier rule to catch all of these automatically.

PR #385, opened 2018-12-12
refactor(wire-protocol): introduce wire protocol `query` method

mbroadst:
NODE-1787

### DIFF
--- a/lib/cursor.js
+++ b/lib/cursor.js
@@ -638,19 +638,6 @@ function initializeCursor(cursor, callback) {
       return callback(new MongoError(`server ${cursor.server.name} does not support collation`));
     }
 
-    cursor.query = cursor.server.wireProtocolHandler.command(
-      cursor.bson,
-      cursor.ns,
-      cursor.cmd,
-      cursor.cursorState,
-      cursor.topology,
-      cursor.options
-    );
-
-    if (cursor.query instanceof MongoError) {
-      return callback(cursor.query);
-    }
-
     function done() {
       if (
         cursor.cursorState.cursorId &&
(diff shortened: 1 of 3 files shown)


lib/cursor.js
```diff
@@ -788,7 +730,48 @@ function initializeCursor(cursor, callback) {
 
       // Return callback
       done(result);
-    });
+    };
+
+    if (cursor.cmd.find != null) {
```
mbroadst: Beware reader, you may find this distasteful (I sure do!), but do not despair. This check was originally done at the `WireProtocol` level, so this merely pushes this anti-pattern closer to the surface and makes it very clear that redesign of the command/operation layer is required.

Much like disambiguating `query` from `command` is the right thing to do, cursor should not act upon all commands equally. A command cursor should use `command`, and a find cursor should use `query`, but none of this is possible for some time so this will have to do for now.

lib/wireprotocol/3_2_support.js
```diff
@@ -151,11 +151,8 @@ class WireProtocol {
     connection.write(query, queryOptions, queryCallback);
   }
 
-  command(bson, ns, cmd, cursorState, topology, options) {
+  query(bson, ns, cmd, cursorState, topology, options) {
     options = options || {};
-    const wireProtocolCommand =
```
kvwalker: Why did we get rid of this?

mbroadst:
@daprahamian @kvwalker this one is up next, it has been rebased on master and ready for review.

lib/cursor.js
daprahamian: Is it worth doing:

```
cursor.query = cursor.server.wireProtocolHandler[cursor.cmd.find != null ? 'query' : 'command'](
         cursor.bson,
         cursor.ns,
         cursor.cmd,
         cursor.cursorState,
         cursor.topology,
         cursor.options
);
```

?

mbroadst: I kind of like that it's big and ugly and repetitive, it makes you _want_ to fix it. Rather than fix it here though, I propose we keep pushing it up the stack so that eventually the correct cursor is using the correct methods (`CommandCursor`, `AggregateCursor` both use `command`, regular `Cursor` defaults to `query`)

mbroadst: It could even be that `Cursor` has a `method` property that defaults to `query`, and the `CommandCursor`, and `AggregateCursor` both pass in `{ method: command }`

PR #386, opened 2018-12-12
refactor(wire-protocol): `command` and `query` now take callbacks

mbroadst:
This completes the work to make a semi-consistent api across all methods on `WireProtocol`, opening up the door for even further consolidation of the codebase here (perhaps `WireProtocolBase`, and switching between `OP_QUERY` and `OP_MSG`)

### DIFF
--- a/lib/cursor.js
+++ b/lib/cursor.js
@@ -7,7 +7,6 @@ const MongoNetworkError = require('./error').MongoNetworkError;
 const mongoErrorContextSymbol = require('./error').mongoErrorContextSymbol;
 const f = require('util').format;
 const collationNotSupported = require('./utils').collationNotSupported;
-const applyCommonQueryOptions = require('./wireprotocol/shared').applyCommonQueryOptions;
 
 const BSON = retrieveBSON();
 const Long = BSON.Long;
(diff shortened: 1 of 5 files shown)


lib/cursor.js
```diff
@@ -105,7 +105,8 @@ var Cursor = function(bson, ns, cmd, options, topology, topologyOptions) {
     batchSize: options.batchSize || cmd.batchSize || 1000,
     currentLimit: 0,
     // Result field name if not a cursor (contains the array of results)
-    transforms: options.transforms
+    transforms: options.transforms,
+    raw: options.raw || (typeof cmd === 'object' && cmd.raw)
```
daprahamian: `typeof null === 'object'`, so we should change this to:

```js
raw: options.raw || (cmd && cmd.raw)
```

lib/topologies/server.js
```diff
@@ -755,33 +755,8 @@ Server.prototype.command = function(ns, cmd, options, callback) {
...
-  // Write options
-  var writeOptions = {
-    raw: typeof options.raw === 'boolean' ? options.raw : false,
-    promoteLongs: typeof options.promoteLongs === 'boolean' ? options.promoteLongs : true,
-    promoteValues: typeof options.promoteValues === 'boolean' ? options.promoteValues : true,
-    promoteBuffers: typeof options.promoteBuffers === 'boolean' ? options.promoteBuffers : false,
-    command: true,
-    monitoring: typeof options.monitoring === 'boolean' ? options.monitoring : false,
-    fullResult: typeof options.fullResult === 'boolean' ? options.fullResult : false,
-    requestId: query.requestId,
```
kvwalker: Why was `requestId` dropped?

lib/cursor.js
mbroadst: changed in prior PRs, rebased into this one.

lib/topologies/server.js
mbroadst: It's hard to see it in this PR, but `requestId` doesn't need to be passed into the options for `Pool.prototype.write` because we are passing the `Query` instance in which already has an accessor for its `requestId` (the one being used in this line, in fact)

PR #387, opened 2018-12-13
fix(mongos-replset): pass connect options to child server instances

mbroadst:
This allows for options determined during uri parsing to make it
down to the creation of child Server instances using the legacy
SDAM topology classes.

NODE-1798

### DIFF
--- a/lib/topologies/mongos.js
+++ b/lib/topologies/mongos.js
@@ -286,7 +286,7 @@ Mongos.prototype.connect = function(options) {
   // Create server instances
   var servers = this.s.seedlist.map(function(x) {
     const server = new Server(
-      Object.assign({}, self.s.options, x, {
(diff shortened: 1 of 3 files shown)


lib/topologies/mongos.js
```diff
@@ -286,7 +286,7 @@ Mongos.prototype.connect = function(options) {
   // Create server instances
   var servers = this.s.seedlist.map(function(x) {
     const server = new Server(
-      Object.assign({}, self.s.options, x, {
+      Object.assign({}, self.s.options, x, options, {
```
daprahamian: I had assumed we were purposefully not propagating the options down here. Is it possible this will have side effects?

mbroadst: spoke about this offline, resolving

PR #395, opened 2019-01-26
refactor(wire-protocol): only single commands are written to pool

mbroadst:
For legacy reasons, `Pool.write` takes an array of commands even
though the driver doesn't actually do anything with an array. It's
easier for the future to accurately reflect what we are passing
around.

NODE-1844

### DIFF
--- a/lib/connection/pool.js
+++ b/lib/connection/pool.js
@@ -11,6 +11,7 @@ const f = require('util').format;
 const Query = require('./commands').Query;
 const CommandResult = require('./command_result');
 const MESSAGE_HEADER_SIZE = require('../wireprotocol/shared').MESSAGE_HEADER_SIZE;
+const COMPRESSION_DETAILS_SIZE = require('../wireprotocol/shared').COMPRESSION_DETAILS_SIZE;
 const opcodes = require('../wireprotocol/shared').opcodes;
 const compress = require('../wireprotocol/compression').compress;
 const compressorIDs = require('../wireprotocol/compression').compressorIDs;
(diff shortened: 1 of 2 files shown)


lib/connection/pool.js
```diff
@@ -1036,61 +1036,49 @@ Pool.prototype.destroy = function(force) {
...
-      serializeCommands(self, commands, result, callback);
-    });
-  } else {
-    // Push the serialization of the command onto results
-    result.push(originalCommandBuffer);
-
-    // Continue recursing through the commands array
-    serializeCommands(self, commands, result, callback);
+  const shouldCompress = !!self.options.agreedCompressor;
+  if (!shouldCompress || !canCompress(command)) {
```
daprahamian: nit: DeMorgan

lib/connection/pool.js
```diff
@@ -1036,61 +1036,49 @@ Pool.prototype.destroy = function(force) {
...
+  // Extract information needed for OP_COMPRESSED from the uncompressed message
+  const originalCommandOpCode = concatenatedOriginalCommandBuffer.readInt32LE(12);
+
+  // Compress the message body
+  compress(self, messageToBeCompressed, function(err, compressedMessage) {
+    if (err) return callback(err, null);
+
+    // Create the msgHeader of OP_COMPRESSED
+    const msgHeader = Buffer.alloc(MESSAGE_HEADER_SIZE);
+    msgHeader.writeInt32LE(MESSAGE_HEADER_SIZE + 9 + compressedMessage.length, 0); // messageLength
```
daprahamian: nit: It would be nice, since we are here, to get rid of the magic #s

PR #396, opened 2019-01-28
NODE-1845: allow connection to be shared with mock server

mbroadst:
Primary goals for this work (from jira ticket):
- Decouple Connection from the Pool by not passing in a messageHandler to the constructor, but use a message event instead
- Move the "connect" logic for creating and initially connection connections out of Connection into its own method, a la net.connect

### DIFF
--- a/lib/connection/connect.js
+++ b/lib/connection/connect.js
@@ -1,0 +1,172 @@
+'use strict';
+const net = require('net');
+const tls = require('tls');
+const Connection = require('./connection');
+
+function connect(messageHandler, options, callback) {
+  if (options.family !== void 0) {
+    makeConnection(options.family, options, (err, socket) => {
+      if (err) {
+        return callback(err, socket); // in the error case, `socket` is the originating error event
+      }
+
+      callback(null, new Connection(messageHandler, socket, options));
+    });
+
+    return;
+  }
+
+  return makeConnection(6, options, (err, ipv6Socket) => {
+    if (err) {
+      // if (this.logger.isDebug()) {
+      //   this.logger.debug(
+      //     `connection ${this.id} for [${this.address}] errored out with [${JSON.stringify(err)}]`
+      //   );
+      // }
+
+      makeConnection(4, options, (err, ipv4Socket) => {
+        if (err) {
+          return callback(err, ipv4Socket); // in the error case, `ipv4Socket` is the originating error event
+        }
+
+        callback(null, new Connection(messageHandler, ipv4Socket, options));
+      });
+
+      return;
+    }
+
+    callback(null, new Connection(messageHandler, ipv6Socket, options));
+  });
+}
+
+const LEGAL_SSL_SOCKET_OPTIONS = [
+  'pfx',
+  'key',
+  'passphrase',
+  'cert',
+  'ca',
+  'ciphers',
+  'NPNProtocols',
+  'ALPNProtocols',
+  'servername',
+  'ecdhCurve',
+  'secureProtocol',
+  'secureContext',
+  'session',
+  'minDHSize'
+];
+
+function parseConnectOptions(family, options) {
+  const host = typeof options.host === 'string' ? options.host : 'localhost';
+  if (host.indexOf('/') !== -1) {
+    return { path: host };
+  }
+
+  const result = {
+    family,
+    host,
+    port: typeof options.port === 'number' ? options.port : 27017,
+    rejectUnauthorized: false
+  };
+
+  return result;
+}
+
+function parseSslOptions(family, options) {
+  const result = parseConnectOptions(family, options);
+
+  // Merge in valid SSL options
+  for (const name in options) {
+    if (options[name] != null && LEGAL_SSL_SOCKET_OPTIONS.indexOf(name) !== -1) {
+      result[name] = options[name];
+    }
+  }
+  // Set options for ssl
+  if (options.ca) result.ca = options.ca;
+  if (options.crl) result.crl = options.crl;
+  if (options.cert) result.cert = options.cert;
+  if (options.key) result.key = options.key;
+  if (options.passphrase) result.passphrase = options.passphrase;
+
+  // Override checkServerIdentity behavior
+  if (options.checkServerIdentity === false) {
+    // Skip the identiy check by retuning undefined as per node documents
+    // https://nodejs.org/api/tls.html#tls_tls_connect_options_callback
+    result.checkServerIdentity = function() {
+      return undefined;
+    };
+  } else if (typeof options.checkServerIdentity === 'function') {
+    result.checkServerIdentity = options.checkServerIdentity;
+  }
+
+  // Set default sni servername to be the same as host
+  if (result.servername == null) {
+    result.servername = result.host;
+  }
+
+  if (typeof options.rejectUnauthorized === 'boolean') {
+    result.rejectUnauthorized = options.rejectUnauthorized;
+  }
+
+  return result;
+}
+
+function makeConnection(family, options, callback) {
+  const useSsl = typeof options.ssl === 'boolean' ? options.ssl : false;
+  const keepAlive = typeof options.keepAlive === 'boolean' ? options.keepAlive : true;
+  let keepAliveInitialDelay =
+    typeof options.keepAliveInitialDelay === 'number' ? options.keepAliveInitialDelay : 300000;
+  const noDelay = typeof options.noDelay === 'boolean' ? options.noDelay : true;
+  const connectionTimeout =
+    typeof options.connectionTimeout === 'number' ? options.connectionTimeout : 30000;
+  const socketTimeout = typeof options.socketTimeout === 'number' ? options.socketTimeout : 360000;
+  const rejectUnauthorized =
+    typeof options.rejectUnauthorized === 'boolean' ? options.rejectUnauthorized : true;
+
+  if (keepAliveInitialDelay > socketTimeout) {
+    keepAliveInitialDelay = Math.round(socketTimeout / 2);
+  }
+
+  let socket;
+  try {
+    if (useSsl) {
+      socket = tls.connect(parseSslOptions(family, options));
+    } else {
+      socket = net.createConnection(parseConnectOptions(family, options));
+    }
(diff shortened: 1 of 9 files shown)


lib/connection/pool.js
```diff
@@ -747,11 +702,11 @@ Pool.prototype.connect = function() {
     }
 
     // attach event handlers
-    connection.once('close', connectionFailureHandler(self, 'close'));
-    connection.once('error', connectionFailureHandler(self, 'error'));
-    connection.once('timeout', connectionFailureHandler(self, 'timeout'));
-    connection.once('parseError', connectionFailureHandler(self, 'parseError'));
-    connection.on('message', self.messageHandler);
+    connection.on('error', self._connectionErrorHandler);
```
daprahamian: why are these switching from `once` to `on`?

lib/connection/pool.js
```diff
@@ -627,13 +647,19 @@ Pool.prototype.socketCount = function() {
   // + this.connectingConnections.length;
 };
 
+function totalConnectionCount(pool) {
```
daprahamian: lets make this a getter on the pool

lib/connection/connect.js
```diff
@@ -0,0 +1,172 @@
...
+  const result = parseConnectOptions(family, options);
+
+  // Merge in valid SSL options
+  for (const name in options) {
+    if (options[name] != null && LEGAL_SSL_SOCKET_OPTIONS.indexOf(name) !== -1) {
+      result[name] = options[name];
+    }
+  }
+  // Set options for ssl
+  if (options.ca) result.ca = options.ca;
```
daprahamian: I'd like to add a comment to investigate these assignments. I think they are probably superfluous now that we have the `for` loop above.

lib/connection/connection.js
```diff
@@ -38,175 +35,252 @@ var debugFields = [
...
+
+    this.port = options.port || 27017;
+    this.host = options.host || 'localhost';
+    this.socketTimeout = typeof options.socketTimeout === 'number' ? options.socketTimeout : 360000;
+
+    // These values are inspected directly in tests, but maybe not necessary to keep around
+    this.keepAlive = typeof options.keepAlive === 'boolean' ? options.keepAlive : true;
+    this.keepAliveInitialDelay =
+      typeof options.keepAliveInitialDelay === 'number' ? options.keepAliveInitialDelay : 300000;
+    this.socketTimeout = typeof options.socketTimeout === 'number' ? options.socketTimeout : 360000;
```
kvwalker: `this.socketTimeout` is also set in https://github.com/mongodb-js/mongodb-core/pull/396/files#diff-b472538a81cdbf2a40c0d2ccd5a7521dR94

lib/connection/pool.js
```diff
@@ -700,87 +683,73 @@ Pool.prototype.connect = function() {
     throw new MongoError('connection in unlawful state ' + this.state);
   }
 
-  var self = this;
-  // Transition to connecting state
+  const self = this;
```
kvwalker: If we're replacing `self` elsewhere, can we use a more descriptive name here too?

kvwalker: Okay, that works for me!

lib/connection/connect.js
```diff
@@ -0,0 +1,167 @@
...
+
+      callback(null, new Connection(socket, options));
+    });
+
+    return;
+  }
+
+  return makeConnection(6, options, (err, ipv6Socket) => {
+    if (err) {
+      // if (this.logger.isDebug()) {
```
kvwalker: Why is this commented out?

lib/connection/connect.js
```diff
@@ -0,0 +1,167 @@
...
+    result.checkServerIdentity = options.checkServerIdentity;
+  }
+
+  // Set default sni servername to be the same as host
+  if (result.servername == null) {
+    result.servername = result.host;
+  }
+
+  if (typeof options.rejectUnauthorized === 'boolean') {
+    result.rejectUnauthorized = options.rejectUnauthorized;
```
kvwalker: Can we add `rejectUnauthorized` to `LEGAL_SSL_SOCKET_OPTIONS`?

lib/connection/connection.js
```diff
@@ -38,175 +35,251 @@ var debugFields = [
...
+   *
+   * @param {string} options.host The server host
+   * @param {number} options.port The server port
+   * @param {number} [options.family=null] IP version for DNS lookup, passed down to Node's [`dns.lookup()` function](https://nodejs.org/api/dns.html#dns_dns_lookup_hostname_options_callback). If set to `6`, will only look for ipv6 addresses.
+   * @param {boolean} [options.keepAlive=true] TCP Connection keep alive enabled
+   * @param {number} [options.keepAliveInitialDelay=300000] Initial delay before TCP keep alive enabled
+   * @param {boolean} [options.noDelay=true] TCP Connection no delay
+   * @param {number} [options.connectionTimeout=30000] TCP Connection timeout setting
+   * @param {number} [options.socketTimeout=360000] TCP Socket timeout setting
+   * @param {boolean} [options.singleBufferSerializtion=true] Serialize into single buffer, trade of peak memory for serialization speed
```
kvwalker: Should be `singleBufferSerialization`.

lib/connection/connection.js
```diff
@@ -38,175 +35,251 @@ var debugFields = [
...
+   * @param {boolean} [options.ssl=false] Use SSL for connection
```
kvwalker: Some of these defaults, like `ssl` and `checkServerIdentity` don't appear to be set here. Where are they set?

lib/connection/connect.js
mbroadst: good catch, added debug calls to `connect` call site to preserve this info

lib/connection/connect.js
mbroadst: yep! added.

lib/connection/connection.js
mbroadst: hah yeah, _should_ be removed! This option is no longer (actually has not been in some time) supported. Removed.

lib/connection/connection.js
mbroadst: Improved these docs with latest patch

lib/connection/connect.js
```diff
@@ -0,0 +1,158 @@
+'use strict';
+const net = require('net');
+const tls = require('tls');
+const Connection = require('./connection');
+
+function connect(options, callback) {
+  if (options.family !== void 0) {
+    makeConnection(options.family, options, (err, socket) => {
```
daprahamian: tiny nit: is there a reason the `return`s are on separate lines? Are we trying to explicitly prevent a return?

lib/connection/connection.js
```diff
@@ -1,25 +1,22 @@
...
+const BinMsg = require('./msg').BinMsg;
+const MongoNetworkError = require('../error').MongoNetworkError;
+const MongoError = require('../error').MongoError;
+const Logger = require('./logger');
+const OP_COMPRESSED = require('../wireprotocol/shared').opcodes.OP_COMPRESSED;
+const OP_MSG = require('../wireprotocol/shared').opcodes.OP_MSG;
+const MESSAGE_HEADER_SIZE = require('../wireprotocol/shared').MESSAGE_HEADER_SIZE;
+const Buffer = require('safe-buffer').Buffer;
+
+let _id = 0;
```
daprahamian: In cmap, I would up writing a `counter` generator. Can we maybe write a util:

```js
function* counter(seed) {
  num = seed || 0;
  while(true) {
    yield num;
    num += 1;
  }
}
```

And then leverage it here? I find the `_id`s that we have all over the place a little hard to parse.

(And I might just be looking for an excuse to use generators 😊 )

lib/connection/connection.js
```diff
@@ -40,175 +36,242 @@ var debugFields = [
...
+    if (!options.bson) {
+      throw new TypeError('must pass in valid bson parser');
+    }
+
+    this.id = _id++;
+    this.options = options;
+    this.logger = Logger('Connection', options);
+    this.bson = options.bson;
+    this.tag = options.tag;
+    this.maxBsonMessageSize = options.maxBsonMessageSize || 1024 * 1024 * 16 * 4;
```
daprahamian: Since we're here: can we make a `DEFAULT_MAX_BSON_MESSAGE_SIZE` const?

lib/connection/connection.js
```diff
@@ -40,175 +36,242 @@ var debugFields = [
...
+
+    this.port = options.port || 27017;
```
daprahamian: makes me really want options management is core :(

lib/connection/pool.js
```diff
@@ -158,6 +158,25 @@ var Pool = function(topology, options) {
   this.numberOfConsecutiveTimeouts = 0;
   // Current pool Index
   this.connectionIndex = 0;
+
+  // event handlers
+  const self = this;
+  this._messageHandler = messageHandler(this);
+  this._connectionCloseHandler = function(err) {
+    connectionFailureHandler(self, 'close', err, this);
```
daprahamian: I'm worried about getting confused about this in the future. Could we maybe do something along the lines of:

```js
const pool = this;
...
this._connectionCloseHandler = function(err) {
  const connection = this;
  connectionFailureHandler(pool, 'close', err, connection);
}
```

or is that overkill?

lib/connection/connect.js
mbroadst: I was inconsistent here, but it's also code copied from `connection.js`. I've been trying to keep `return` on its own line especially in cases where we're returning a function call, in order to make it very clear that execution stops _here_. I'll change these missed cases.

lib/connection/connection.js
mbroadst: agreed. _soon_

lib/connection/pool.js
mbroadst: sure

lib/connection/connection.js
mbroadst: I'd prefer to leave this alone if that's alright with you. Your solution in the new pool refactor is preferred, and I'd like to just leave the old code here alone until we move that way.

PR #399, opened 2019-01-30
NODE-1741: implement convenient transactions api (core)

mbroadst:
This is the core component (the meat) of NODE-1741. There were some design decisions made here, most notably that only a `Promise` may be returned from the lambda passed into `withTransaction`. Outside of that, this mostly follows the spec to the letter.

### DIFF
--- a/lib/sessions.js
+++ b/lib/sessions.js
@@ -265,10 +265,11 @@ class ClientSession extends EventEmitter {
    */
   withTransaction(fn, options) {
     const startTime = Date.now();
-    return retryTransaction(this, startTime, fn, options);
+    return attemptTransaction(this, startTime, fn, options);
   }
 }
 
+const MAX_WITH_TRANSACTION_TIMEOUT = 120000;
 const WRITE_CONCERN_FAILED_CODE = 64;
 const UNSATISFIABLE_WRITE_CONCERN_CODE = 100;
 const UNKNOWN_REPL_WRITE_CONCERN_CODE = 79;
@@ -277,6 +278,10 @@ const NON_DETERMINISTIC_WRITE_CONCERN_ERRORS = [
   'UnknownReplWriteConcern',
   'UnsatisfiableWriteConcern'
 ];
+
+function hasNotTimedOut(startTime, max) {
+  return Date.now() - startTime < max;
+}
 
 function isWriteConcernTimeoutError(err) {
   return err.code === WRITE_CONCERN_FAILED_CODE && !!(err.errInfo && err.errInfo.wtimeout === true);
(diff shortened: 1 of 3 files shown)


lib/sessions.js
```diff
@@ -243,6 +243,106 @@ class ClientSession extends EventEmitter {
...
+      self.startTransaction(options);
+
+      // TODO: should we support callbacks?
+      return fn(self)
+        .then(() => {
+          if (
+            [
+              TxnState.NO_TRANSACTION,
+              TxnState.TRANSACTION_COMMITTED,
+              TxnState.TRANSACTION_ABORTED
```
daprahamian: a) hoist the array out into a const `Set`
b) maybe make a small function that wraps this so we know what it is doing.

lib/sessions.js
```diff
@@ -243,6 +243,106 @@ class ClientSession extends EventEmitter {
...
+   * return a Promise will result in undefined behavior.
+   *
+   * @param {WithTransactionCallback} fn
+   * @param {TransactionOptions} [options] Optional settings for the transaction
+   */
+  withTransaction(fn, options) {
+    const self = this;
+    const startTime = Date.now();
+
+    function retryCommit() {
```
daprahamian: Names like `retryCommit` and `retryTransaction` make this code confusing when they are being invoked for the first time. Maybe `commitTransaction` and `runTransaction` would be better.

lib/sessions.js
```diff
@@ -243,6 +243,106 @@ class ClientSession extends EventEmitter {
...
+      return self.commitTransaction().catch(err => {
+        if (
+          !isWriteConcernTimeoutError(err) &&
+          (err instanceof MongoError && err.hasErrorLabel('UnknownTransactionCommitResult')) &&
+          Date.now() - startTime < 120000
```
daprahamian: magic # should be a const somewhere, with a comment block explaining why we are going with 2 minutes.

lib/sessions.js
```diff
@@ -243,6 +243,106 @@ class ClientSession extends EventEmitter {
...
+            ].indexOf(self.transaction.state) !== -1
+          ) {
+            // Assume user provided function intentionally ended the transaction
+            return;
+          }
+
+          return retryCommit();
+        })
+        .catch(err => {
+          function maybeRetryOrThrow(err) {
```
daprahamian: lets hoist this out of being defined in the callback.

lib/sessions.js
```diff
@@ -243,6 +243,106 @@ class ClientSession extends EventEmitter {
...
+  }
+}
+
+function isWriteConcernTimeoutError(err) {
+  return err.code === 64 && !!(err.errInfo && err.errInfo.wtimeout === true);
+}
+
+function isUnknownTransactionCommitResult(err) {
+  return (
+    ['CannotSatisfyWriteConcern', 'UnknownReplWriteConcern', 'UnsatisfiableWriteConcern'].indexOf(
```
daprahamian: hoist label list into `const`. Or even better, into a `Set`

lib/sessions.js
```diff
@@ -243,6 +243,106 @@ class ClientSession extends EventEmitter {
...
+          return maybeRetryOrThrow(err);
+        });
+    }
+
+    return retryTransaction();
+  }
+}
+
+function isWriteConcernTimeoutError(err) {
+  return err.code === 64 && !!(err.errInfo && err.errInfo.wtimeout === true);
```
daprahamian: Magic # 64

lib/sessions.js
daprahamian: Also, maybe instead of doing `Date.now() - startTime < 120000` all over the place, we can instead do `const timeout = Date.now() + 120000`, and then just check `Date.now() >= timeout`

lib/sessions.js
```diff
@@ -243,6 +243,106 @@ class ClientSession extends EventEmitter {
...
+
+        throw err;
+      });
+    }
+
+    function retryTransaction() {
+      self.startTransaction(options);
+
+      // TODO: should we support callbacks?
+      return fn(self)
```
daprahamian: If someone forgets to return a promise, this is gonna lead to a nasty looking bug. The way I see it, we have two alternatives:

- use `Promise.resolve().then(() => fn(self))`: This will ensure that we are dealing with a promise, but might make it harder for a user to see why their transaction failed.
- inspect the return value of `fn`, throw an explicit driver error if the ret is not a promise.

daprahamian: Also we should NOT support callbacks.

lib/sessions.js
mbroadst: made `userExplicitlyEndedTransaction`

lib/sessions.js
mbroadst: renamed to `attemptTransaction` and `attemptTransactionCommit`

lib/sessions.js
mbroadst: 120000 is indeed a magic number that will go away in a few months, in favor of a client-side operations timeout. I'm inclined to just leave this as-is at the moment.

lib/sessions.js
mbroadst: What's the perceived benefit here? The function already defines what we're checking for (these error labels tell us whether or not an error is an UnknownTransactionCommit), and v8 will optimize this function given that the input params don't change shape.

lib/sessions.js
mbroadst: added `WRITE_CONCERN_FAILED_CODE`

lib/sessions.js
mbroadst: added `hasNotTimedOut` and `MAX_WITH_TRANSACTION_TIMEOUT`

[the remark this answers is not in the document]
mbroadst: `added hasNotTimedOut and MAX_WITH_TRANSACTION_TIMEOUT`

lib/sessions.js
mbroadst: I left this one alone, it was the odd duck out and IMO not useful as a package level function.

lib/sessions.js
mbroadst: Removed comment, added inspection of return type + a test in the native portion of this PR.

mbroadst:
@daprahamian @kvwalker all issues should be addressed, PTAL

lib/sessions.js
```diff
@@ -243,6 +244,130 @@ class ClientSession extends EventEmitter {
...
+  return Date.now() - startTime < max;
+}
+
+function isWriteConcernTimeoutError(err) {
+  return err.code === WRITE_CONCERN_FAILED_CODE && !!(err.errInfo && err.errInfo.wtimeout === true);
+}
+
+function isUnknownTransactionCommitResult(err) {
+  return (
+    NON_DETERMINISTIC_WRITE_CONCERN_ERRORS.indexOf(err.codeName) === -1 &&
```
daprahamian: To follow up on https://github.com/mongodb-js/mongodb-core/pull/399#discussion_r253275147:

It is more semantically correct to use a `Set` over an `Array` when checking if something exists. It is clearer to have a `Set` and say `!set.has(item)` then it is to do an `indexOf` check (as well as being less verbose)

As for hoisting it out, it is more a verbosity issue than a performance issue. It is clearer to see:

```
NON_DETERMINISTIC_WRITE_CONCERN_ERRORS.has(err.codeName);
```

instead of an inline list. And should we add more error conditions, it will be more maintainable.

Overall, I'd like to stop using Arrays when we could be using Sets.

mbroadst: 👍 switched to using `Set`

PR #400, opened 2019-02-01
fix(uri_parser): fix normalizing option names in connection string

kvwalker:
Fixes NODE-1658

### DIFF
--- a/lib/uri_parser.js
+++ b/lib/uri_parser.js
@@ -11,6 +11,48 @@ const ReadPreference = require('./topologies/read_preference');
  */
 const HOSTS_RX = /(mongodb(?:\+srv|)):\/\/(?: (?:[^:]*) (?: : ([^@]*) )? @ )?([^/?]*)(?:\/|)(.*)/;
 
+const URI_OPTIONS = {
+  appname: { type: 'string' },
+  authmechanism: { type: 'array' },
+  authmechanismproperties: { type: 'array' },
+  authsource: { type: 'string' },
+  compressors: { type: 'array' },
+  connecttimeoutms: { type: 'number' },
+  fsync: { type: 'boolean' },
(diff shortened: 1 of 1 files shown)


lib/uri_parser.js
```diff
@@ -362,15 +362,50 @@ function applyAuthExpectations(parsed) {
...
 function parseQueryString(query, options) {
   const result = {};
-  let parsedQueryString = qs.parse(query);
+  let splitQueryString = query.split('&');
 
-  for (const key in parsedQueryString) {
-    const value = parsedQueryString[key];
-    if (value === '' || value == null) {
-      throw new MongoParseError('Incomplete key value pair for option');
+  let queryKeys = [];
```
daprahamian: is it necessary to use arrays for these? It seems like we could achieve the results we want by using a `Map<normalizedKey, value>` and a `Set<duplicateKeys>`

daprahamian: Also, does this properly handle values that are supposed to be array values?

kvwalker: Good catch. This doesn't work for handling values that are supposed to be array values if they're passed in like `foo=bar&foo=baz`. I'm not sure how to handle a situation where we have something like `readPreference=primary&foo=bar&foo=baz&readPreference=secondary` where `readPreference` should be `secondary`, but `foo` should be `[bar, baz]`...

daprahamian: I think the best route would be to break it into the following processes:

1. parse every option into a `Map<normalizedKey, values[]>`.
2. for every `key, values[]` , run validation on the values passed in. This would be a good place to warn if a values that doesn't support arrays has multiple values.

I think step 2 could probably be accomplished by refactoring `parseQueryStringItemValue` and `applyConnectionStringOption`. Both of them do some sort of validation / type checking.

PR #307, opened 2018-05-09
use async randomBytes()

jcallu:
coming back around to retry a PR for this.

PR #402, opened 2019-02-05
refactor(auth): switch to MongoCredentials and cleanup

mbroadst:
A major refactor to rewrite core to use MongoCredentials objects
instead of individually passing around auth fields. Also removes
significant amounts of duplicate code in auth providers. Also
switches to the async version of crypto.randomBytes to
resolve #307.

Fixes NODE-1436
Fixes NODE-1442

### DIFF
--- a/lib/auth/mongocr.js
+++ b/lib/auth/mongocr.js
@@ -1,192 +1,51 @@
 'use strict';
 
-var f = require('util').format,
-  crypto = require('crypto'),
-  MongoError = require('../error').MongoError;
-
-var AuthSession = function(db, username, password) {
-  this.db = db;
-  this.username = username;
-  this.password = password;
-};
-
-AuthSession.prototype.equal = function(session) {
-  return (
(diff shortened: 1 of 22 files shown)


lib/auth/mongocr.js
```diff
@@ -1,192 +1,54 @@
...
+   * Implementation of authentication for a single connection
+   * @override
+   */
+  _authenticateSingleConnection(sendAuthCommand, connection, credentials, callback) {
+    const username = credentials.username;
+    const password = credentials.password;
+    const source = credentials.source;
+
+    // Write the commmand on the connection
+    const getNonceCommand = this._buildCommand(`${source}.$cmd`, { getnonce: 1 });
```
kvwalker: Looks like this is causing test failures because `_buildCommand` isn't a function.

mbroadst: thanks, I forgot to push that change

PR #403, opened 2019-02-11
NODE-1085: implement OP_MSG

mbroadst:
This builds on @daprahamian 's original work, and brings us to full `OP_MSG` support on server versions that support the protocol. As noted previously, the way we are currently getting around mock server tests is by introduction of a new `__nodejs_mock_server__` ismaster field provided in v1.0.1 of the mock server.


lib/connection/msg.js
```diff
@@ -0,0 +1,293 @@
...
+const opcodes = require('../wireprotocol/shared').opcodes;
+const databaseNamespace = require('../wireprotocol/shared').databaseNamespace;
+
+// Incrementing request id
+let _requestId = 0;
+
+// Msg Flags
+const OPTS_CHECKSUM_PRESENT = 1;
+const OPTS_MORE_TO_COME = 2;
+const OPTS_EXHAUST_ALLOWED = 1 >> 16;
```
daprahamian: This seems like a bug. Shouldn't it be `1 << 16`?

lib/connection/msg.js
```diff
@@ -0,0 +1,293 @@
...
+
+      const doc = this.bson.deserialize(this.documents[0], _options);
+      this.documents = [doc];
+    }
+
+    this.parsed = true;
+  }
+}
+
+Msg.flags = {
```
daprahamian: These are duplicated above with `OPTS_*`. We should pick one and stick with it.

lib/connection/msg.js
```diff
@@ -0,0 +1,293 @@
...
+  CHECKSUM_PRESENT: 1,
+  MORE_TO_COME: 2,
+  EXHAUST_ALLOWED: 1 << 16
+};
+
+function writeInt32ListToUint8Buffer(buffer, int32List, start) {
+  let index = start || 0;
+  int32List.forEach(int32 => {
+    buffer[index] = int32 & 0xff;
+    buffer[index + 1] = (int32 >> 8) & 0xff;
```
daprahamian: Is it worth it to make these `>>>` for the explicit cast to `uint32`?

lib/connection/msg.js
```diff
@@ -0,0 +1,293 @@
...
+    this.command.$db = databaseNamespace(ns);
+
+    // Ensure empty options
+    this.options = options || {};
+
+    // Additional options
+    this.requestId = Msg.getRequestId();
+
+    // Serialization option
+    this.serializeFunctions =
```
kvwalker: Can we document these options?

lib/connection/msg.js
```diff
@@ -0,0 +1,293 @@
...
+  }
+}
+
+Msg.getRequestId = function() {
+  return ++_requestId;
+};
+
+class BinMsg {
+  constructor(bson, message, msgHeader, msgBody, opts) {
+    opts = opts || { promoteLongs: true, promoteValues: true, promoteBuffers: false };
```
kvwalker: Can we document these options?

lib/connection/msg.js
mbroadst: nice catch, I didn't touch your implementation much, as I imagine there will be some fixes and refactoring done in upcoming projects (FLE, exhaust cursors, etc)

lib/connection/msg.js
mbroadst: 👍 removed in favor of private constants (seems more consistent with the current codebase)

lib/connection/msg.js
mbroadst: I've actually always wondered why we don't just use the [built-in buffer methods](https://nodejs.org/api/buffer.html#buffer_buf_writeuint32be_value_offset) for writing this data. Can you think of an argument against it? Otherwise, I'll just quickly convert these over.

lib/connection/msg.js
mbroadst: should we document these everywhere? They are just options eventually passed down to BSON for deserialization. I think we've adequately documented them at the top level, and don't need to see these docs at every step they are considered.

lib/connection/msg.js
mbroadst: Similarly, these are BSON-related options

lib/connection/msg.js
kvwalker: That's fair. My thought process was that if we need to do anything similar to options management again, it's easiest to see where options are used by scanning through the documentation. That being said, I only care 1/10 on this so I'm fine with not documenting these.

lib/connection/msg.js
```diff
@@ -93,26 +93,14 @@ class Msg {
...
-      // Create type 1 query
-      totalLength += this.makeSequenceSegment(buffers, argument, command[argument]);
-    } else {
-      totalLength += this.makeDocumentSegment(buffers, command);
-    }
-
-    writeInt32ListToUint8Buffer(header, [totalLength, this.requestId, 0, opcodes.OP_MSG, flags]);
+    totalLength += this.makeDocumentSegment(buffers, command);
 
+    header.writeInt32LE(totalLength, 0); // messageLength
```
daprahamian: Can any of these values be negative? I think to guard ourselves we should make the `writeUInt32LE`?

mbroadst: The spec itself defines them as `int32`. Though I agree I can't imagine why they would be negative I wouldn't want to be non-compliant if they ever decide to do something with negative values here.

PR #404, opened 2019-02-15
fix(sessions): enable sessions in OP_MSG

mbroadst:
Sessions information is not presently applied to OP_MSG messages
because of a patch that was left out of the OP_MSG PR. The real
fix for this is to merge all of the sessions-related code in
our pool into the `command` wire protocol method, but that work
shouldn't hold up sessions support today.


mbroadst:
@daprahamian I'd say its mostly on Travis config + evergreen config, but no worries this was never released

PR #405, opened 2019-02-16
refactor(handshake): pull out initial handshake into `connect`

mbroadst:
This removes the initial handshake code from the middle of the
Server topology implementation into our new `connect` method,
taking us one step closer to isolating all connection
initialization to a single location. The next step will be to
authenticate the connections.

NODE-1863

### DIFF
--- a/lib/connection/connect.js
+++ b/lib/connection/connect.js
@@ -92,6 +92,7 @@ function performInitialHandshake(conn, options, callback) {
 
     if (!isSupportedServer(ismaster)) {
       const latestSupportedVersion = '2.6';
+      const latestSupportedMaxWireVersion = 2;
       const message =
         'Server at ' +
         options.host +
@@ -99,7 +100,9 @@ function performInitialHandshake(conn, options, callback) {
         options.port +
         ' reports wire version ' +
         (ismaster.maxWireVersion || 0) +
-        ', but this version of Node.js Driver requires at least 2 (MongoDB' +
+        ', but this version of Node.js Driver requires at least ' +
+        latestSupportedMaxWireVersion +
+        ' (MongoDB' +
         latestSupportedVersion +
         ').';
 
--- a/lib/connection/connect.js
+++ b/lib/connection/connect.js
@@ -2,6 +2,9 @@
 const net = require('net');
 const tls = require('tls');
 const Connection = require('./connection');
+const Query = require('./commands').Query;
+const createClientInfo = require('../topologies/shared').createClientInfo;
+const MongoError = require('../error').MongoError;
 
 function connect(options, callback) {
   if (options.family !== void 0) {
@@ -11,7 +14,7 @@ function connect(options, callback) {
         return;
       }
 
-      callback(null, new Connection(socket, options));
+      performInitialHandshake(new Connection(socket, options), options, callback);
     });
 
     return;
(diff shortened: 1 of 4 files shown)


lib/connection/connect.js
```diff
@@ -157,4 +253,39 @@ function makeConnection(family, options, callback) {
   socket.once('connect', connectHandler);
 }
 
+const CONNECTION_ERROR_EVENTS = ['error', 'close', 'timeout', 'parseError'];
+function runCommand(conn, ns, command, options, callback) {
```
mbroadst: I expect some discussion around this method. Ideally our wire protocol primitives (`command`, `getMore`, etc) would be placed directly on the `Connection` class, in which case I could just use `conn.command(...)`. Unfortunately, that's not the case today. If there is consensus, I could move _just_ `command` onto `Connection`, since this will be the same method that we use for authentication.

lib/connection/connect.js
```diff
@@ -25,13 +28,106 @@ function connect(options, callback) {
...
 
       return;
     }
 
-    callback(null, new Connection(ipv6Socket, options));
+    performInitialHandshake(new Connection(ipv6Socket, options), options, callback);
+  });
+}
+
+function isSupportedServer(ismaster) {
```
mbroadst: `isSupportedServer`, `getSaslSupportedMechs` are both copied from `server.js` without modification.

lib/connection/connect.js
```diff
@@ -25,13 +28,106 @@ function connect(options, callback) {
...
+    if (!isSupportedServer(ismaster)) {
+      const latestSupportedVersion = '2.6';
+      const message =
+        'Server at ' +
+        options.host +
+        ':' +
+        options.port +
+        ' reports wire version ' +
+        (ismaster.maxWireVersion || 0) +
+        ', but this version of Node.js Driver requires at least 2 (MongoDB' +
```
kvwalker: I'm a little confused by the wording of this error message...should it be `but this version of Node.js Driver requires at least MongoDB + latestSupportedVersion + ').'`? Why does it say `requires at least 2 (MongoDB`?

lib/connection/connect.js
kvwalker: I'd prefer to leave `command` in the wire protocol until we can move all of the wire protocol primitives to `Connection`. I think it would be less organized to only move `command` here.

lib/connection/connect.js
```diff
@@ -25,13 +28,106 @@ function connect(options, callback) {
           return;
         }
 
-        callback(null, new Connection(ipv4Socket, options));
+        performInitialHandshake(new Connection(ipv4Socket, options), options, callback);
```
daprahamian: for DRY purposes:

perhaps:

```
const handshake = (socket) => performInitialHandshake(new Connection(socket, options), options, callback);
```

And then use it where it is necessary.

lib/connection/connect.js
```diff
@@ -25,13 +28,106 @@ function connect(options, callback) {
...
+function isSupportedServer(ismaster) {
+  return ismaster && typeof ismaster.maxWireVersion === 'number' && ismaster.maxWireVersion >= 2;
+}
+
+function getSaslSupportedMechs(options) {
+  if (!(options && options.credentials)) {
+    return {};
+  }
+
+  const credentials = options.credentials;
```
daprahamian: Can we add:

```js
// TODO: revisit whether or not items like `options.user` and `options.dbName` should be checked here
```

lib/connection/connect.js
daprahamian: Agree with @kvwalker: lets not move these over piecemeal.

lib/connection/connect.js
mbroadst: This is supposed to be configurable (see new SDAM implementation), but what it's saying is that the wire version reported from the remote server is not at least the value `2` (MongoDB 2.6). `latestSupportedVersion` is the MongoDB version, what we would need here is the `latestSupportedMaxWireVersion`

mbroadst:
> One item I'd like to double-check is how this will align with CMAP. According to the spec, the Pool needs to be able to create a virtual connection in a non-async manner. Now that we are preventing the creation and return of a connection object until after the handshake, we'd need to find a way to create a virtual connection for the new pool.

Agreed that we might need to tweak some of the CMAP rules for this, but do you generally agree with this approach? Or would you like to discuss design alternatives. Personally, I feel that placing all of this in the `connect` method is a very desirable separation of concerns, but very willing to discuss alternatives!

daprahamian:
> @daprahamian:
> 
> > One item I'd like to double-check is how this will align with CMAP. According to the spec, the Pool needs to be able to create a virtual connection in a non-async manner. Now that we are preventing the creation and return of a connection object until after the handshake, we'd need to find a way to create a virtual connection for the new pool.
> 
> Agreed that we might need to tweak some of the CMAP rules for this, but do you generally agree with this approach? Or would you like to discuss design alternatives. Personally, I feel that placing all of this in the `connect` method is a very desirable separation of concerns, but very willing to discuss alternatives!

I do agree with the approach. I just want to make sure that we do not make it too difficult to incorporate this into CMAP, and that we do it with an awareness that we need to have some sort of "virtual connection" in the future Pool. The only really worrying part about it is that we cannot get a placeholder to `Connection` until, at the earliest, after the tcp/tls connection is established.

lib/connection/connect.js
mbroadst: added

lib/connection/connect.js
mbroadst: I had considered this, but was also trying to reduce the stack trace. Do you feel strongly about this one? It's only repeated thrice (and only ever will be)

lib/connection/connect.js
mbroadst: added `latestSupportedMaxWireVersion`

lib/connection/connect.js
daprahamian: My thoughts on this are only that if we ever change this, we have to change it in 3 places. It's not a dealbreaker for me though.

PR #407, opened 2019-02-19
fix(scram): allow errors to be passed through callbacks

kvwalker:
Fixes NODE-1876

### DIFF
--- a/lib/auth/scram.js
+++ b/lib/auth/scram.js
@@ -143,7 +143,11 @@ class ScramSHA extends AuthProvider {
 
       processedPassword = saslprep ? saslprep(password) : password;
     } else {
-      processedPassword = passwordDigest(username, password);
+      try {
+        processedPassword = passwordDigest(username, password);
+      } catch (e) {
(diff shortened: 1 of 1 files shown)


lib/auth/scram.js
```diff
@@ -143,7 +161,7 @@ class ScramSHA extends AuthProvider {
 
       processedPassword = saslprep ? saslprep(password) : password;
     } else {
-      processedPassword = passwordDigest(username, password);
+      processedPassword = passwordDigest(username, password, callback);
```
daprahamian: rather than changing the `passwordDigest` function, can we instead wrap this in a `try`/`catch`?

```js
} else {
  try {
    processedPassword = passwordDigest(username, password);
  } catch (e) {
    return callback(e);
  }
}
```

mbroadst: I prefer this approach as well, it requires less passing of `callback` around. It's a little bit weird that the proposed approach above would use `callback` to return errors, but otherwise return a value.

PR #409, opened 2019-02-21
fix(command): only add `$query` to command in OP_QUERY

mbroadst:


lib/wireprotocol/command.js
```diff
@@ -24,7 +24,12 @@ function command(server, ns, cmd, options, callback) {
   const readPreference = getReadPreference(cmd, options);
 
   let finalCmd = Object.assign({}, cmd);
-  if (isSharded(server) && readPreference && readPreference.preference !== 'primary') {
+  if (
+    isSharded(server) &&
+    !supportsOpMsg(server) &&
```
daprahamian: nit: pull the `supportsOpMsg` call into a const higher up so it could be reused at line 58

PR #410, opened 2019-02-22
fix(uri_parser): support URI Options spec tests

kvwalker:
Fixes NODE-1740

### DIFF
--- a/lib/uri_parser.js
+++ b/lib/uri_parser.js
@@ -197,10 +197,12 @@ const CASE_TRANSLATION = {
   serverselectiontimeoutms: 'serverSelectionTimeoutMS',
   serverselectiontryonce: 'serverSelectionTryOnce',
   heartbeatfrequencyms: 'heartbeatFrequencyMS',
-  appname: 'appName',
   retrywrites: 'retryWrites',
   uuidrepresentation: 'uuidRepresentation',
-  zlibcompressionlevel: 'zlibCompressionLevel'
+  zlibcompressionlevel: 'zlibCompressionLevel',
+  tlsallowinvalidcertificates: 'tlsAllowInvalidCertificates',
+  tlsallowinvalidhostnames: 'tlsAllowInvalidHostnames',
+  tlsinsecure: 'tlsInsecure'
 };
 
 /**
@@ -225,6 +227,7 @@ function applyConnectionStringOption(obj, key, value, options) {
   } else if (key === 'appname') {
     value = decodeURIComponent(value);
   } else if (key === 'readconcernlevel') {
+    obj['readConcernLevel'] = value;
     key = 'readconcern';
     value = { level: value };
   }
--- a/lib/uri_parser.js
+++ b/lib/uri_parser.js
@@ -202,7 +202,12 @@ const CASE_TRANSLATION = {
   zlibcompressionlevel: 'zlibCompressionLevel',
   tlsallowinvalidcertificates: 'tlsAllowInvalidCertificates',
   tlsallowinvalidhostnames: 'tlsAllowInvalidHostnames',
-  tlsinsecure: 'tlsInsecure'
+  tlsinsecure: 'tlsInsecure',
+  tlscafile: 'tlsCAFile',
+  tlscertificatekeyfile: 'tlsCertificateKeyFile',
+  tlscertificatekeyfilepassword: 'tlsCertificateKeyFilePassword',
+  wtimeout: 'wTimeoutMS',
+  j: 'journal'
 };
 
 /**
@@ -259,17 +264,27 @@ function applyConnectionStringOption(obj, key, value, options) {
     throw new MongoParseError('zlibCompressionLevel must be an integer between -1 and 9');
   }
 
-  // special cases
-  if (key === 'compressors' || key === 'zlibcompressionlevel') {
-    obj.compression = obj.compression || {};
-    obj = obj.compression;
-  }
-
   if (key === 'authmechanismproperties') {
     if (typeof value.SERVICE_NAME === 'string') obj.gssapiServiceName = value.SERVICE_NAME;
     if (typeof value.SERVICE_REALM === 'string') obj.gssapiServiceRealm = value.SERVICE_REALM;
     if (typeof value.CANONICALIZE_HOST_NAME !== 'undefined') {
       obj.gssapiCanonicalizeHostName = value.CANONICALIZE_HOST_NAME;
+    }
+  }
+
+  if (key === 'readpreferencetags') {
+    if (Array.isArray(value)) {
+      const parsedTags = [];
+
+      for (let i = 0; i < value.length; i++) {
+        parsedTags[i] = {};
+        value[i].split(',').forEach(individualTag => {
+          const splitTag = individualTag.split(':');
+          parsedTags[i][splitTag[0]] = splitTag[1];
+        });
+      }
+
+      value = parsedTags;
     }
   }
 
--- a/lib/uri_parser.js
+++ b/lib/uri_parser.js
@@ -273,17 +273,7 @@ function applyConnectionStringOption(obj, key, value, options) {
   }
 
   if (key === 'readpreferencetags' && Array.isArray(value)) {
-    const parsedTags = [];
-
-    for (let i = 0; i < value.length; i++) {
-      parsedTags[i] = {};
-      value[i].split(',').forEach(individualTag => {
-        const splitTag = individualTag.split(':');
-        parsedTags[i][splitTag[0]] = splitTag[1];
-      });
-    }
-
-    value = parsedTags;
+    value = splitArrayOfMultipleReadPreferenceTags(value);
   }
 
   // set the actual value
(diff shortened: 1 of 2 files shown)


lib/uri_parser.js
```diff
@@ -197,10 +197,12 @@ const CASE_TRANSLATION = {
   serverselectiontimeoutms: 'serverSelectionTimeoutMS',
   serverselectiontryonce: 'serverSelectionTryOnce',
   heartbeatfrequencyms: 'heartbeatFrequencyMS',
-  appname: 'appName',
```
kvwalker: The spec specifies that `appname` is all lowercase. This change didn't break any tests, but do we think that users are depending on having `appName`?

lib/uri_parser.js
```diff
@@ -225,6 +227,7 @@ function applyConnectionStringOption(obj, key, value, options) {
   } else if (key === 'appname') {
     value = decodeURIComponent(value);
   } else if (key === 'readconcernlevel') {
+    obj['readConcernLevel'] = value;
```
kvwalker: The spec tests check that `readConcernLevel` is set instead of `readConcern: { level: x }`.

lib/uri_parser.js
daprahamian: if we're going to be non-spec-compliant on this, we should make sure it has a deprecation message.

lib/uri_parser.js
daprahamian: what is happening here?

lib/uri_parser.js
mbroadst: that sounds reasonable (re: deprecation)

lib/uri_parser.js
mbroadst: this does look like it could just be added to `CASE_TRANSLATION` above?

kvwalker: What we currently do with `readConcernLevel` is set `readConcern: { level: value }`. The spec tests check that we have `readConcernLevel: value` so that's why I am pulling the `value` up a level.

lib/uri_parser.js
kvwalker: This change makes us spec compliant. The spec says that `appname` should be lowercase, not camelcase.

daprahamian: Eh, how we represent it internally doesn't matter. I think it will be fine.

lib/uri_parser.js
```diff
@@ -270,6 +272,22 @@ function applyConnectionStringOption(obj, key, value, options) {
     }
   }
 
+  if (key === 'readpreferencetags') {
```
daprahamian: nit: this can be flattened to:

```js
if (key === 'readpreferencetags' && Array.isArray(value)) {
  ...
}
```

lib/uri_parser.js
```diff
@@ -384,6 +404,71 @@ function parseQueryString(query, options) {
...
+    queryStringKeys.indexOf('tlsInsecure') !== -1 &&
+    (queryStringKeys.indexOf('tlsAllowInvalidCertificates') !== -1 ||
+      queryStringKeys.indexOf('tlsAllowInvalidHostnames') !== -1)
+  ) {
+    throw new MongoParseError(
+      'The tlsInsecure option cannot be used with tlsAllowInvalidCertificates or tlsAllowInvalidHostnames.'
+    );
+  }
+
+  if (queryStringKeys.indexOf('tls') !== -1 && queryStringKeys.indexOf('ssl') !== -1) {
```
daprahamian: nit: stick `queryStringKeys.indexOf('tls')` and `queryStringKeys.indexOf('ssl') !== -1` in variables and reuse them.

lib/uri_parser.js
```diff
@@ -270,6 +272,20 @@ function applyConnectionStringOption(obj, key, value, options) {
     }
   }
 
+  if (key === 'readpreferencetags' && Array.isArray(value)) {
+    const parsedTags = [];
+
+    for (let i = 0; i < value.length; i++) {
+      parsedTags[i] = {};
```
mbroadst: this seems ripe for refactoring, can we pull this out into a well-named method and add a unit test or two stressing the implementation out?

lib/uri_parser.js
```diff
@@ -384,6 +402,74 @@ function parseQueryString(query, options) {
...
+ */
+function checkTLSOptions(queryString) {
+  const queryStringKeys = Object.keys(queryString);
+  if (
+    queryStringKeys.indexOf('tlsInsecure') !== -1 &&
+    (queryStringKeys.indexOf('tlsAllowInvalidCertificates') !== -1 ||
+      queryStringKeys.indexOf('tlsAllowInvalidHostnames') !== -1)
+  ) {
+    throw new MongoParseError(
+      'The tlsInsecure option cannot be used with tlsAllowInvalidCertificates or tlsAllowInvalidHostnames.'
```
mbroadst: nit: I think we tend to wrap the actual option names in backticks, or at least quotes?

lib/uri_parser.js
```diff
@@ -384,6 +402,74 @@ function parseQueryString(query, options) {
...
+    );
+  }
+
+  const queryStringHasTLS = queryStringKeys.indexOf('tls') !== -1;
```
mbroadst: starting here it looks like this can all be factored into a method that takes in either `ssl` or `tls` and runs the checks:

```
function checkTlsValues(type) {
  ...
}

checkTlsValues('tls');
checkTlsValues('ssl');
```

lib/uri_parser.js
```diff
@@ -384,6 +402,74 @@ function parseQueryString(query, options) {
...
+}
+
+/**
+ * Checks a query string to ensure all tls/ssl options are the same.
+ *
+ * @param {string} key The key (tls or ssl) to check
+ * @param {string} queryString The query string to check
+ * @throws {MongoParseError}
+ */
+function checkAllOptions(key, queryString) {
```
mbroadst: `checkAllOptions` doesn't seem to describe what's going on here. It seems like this can be rolled into the refactor proposed above, or otherwise maybe we can rename to like `assertTlsOptionsAreEqual` or something

test/tests/unit/connection_string_tests.js
```diff
@@ -100,11 +100,8 @@ describe('Connection String', function() {
       'mongodb://localhost/?compressors=zlib&zlibCompressionLevel=4',
       (err, result) => {
         expect(err).to.not.exist;
-        expect(result.options).to.have.property('compression');
-        expect(result.options.compression).to.eql({
-          compressors: ['zlib'],
-          zlibCompressionLevel: 4
-        });
+        expect(result.options.compressors).to.eql(['zlib']);
```
mbroadst: Will this break anything in native/elsewhere? It looks like we used to verify that `compression` existed, and now we don't. Let's be careful about removing these types of checks

lib/uri_parser.js
mbroadst: additional nit: it seems more like we're "asserting" here rather than just "checking"

test/tests/unit/connection_string_tests.js
kvwalker: Hmm it might since I'm removing L260. Should I support both `compressors` and `compression`?

lib/uri_parser.js
mbroadst: @daprahamian can this be resolved?

test/tests/unit/connection_string_tests.js
mbroadst: If this breaks compression we should be seeing it in unit tests. Like if this does what I was proposing it might, then compression won't work at all. Can you go take a look and verify that it's not doing that?

kvwalker: No unit tests fail. I'm pretty sure this doesn't break compression because tests like [this](https://github.com/mongodb-js/mongodb-core/blob/master/test/tests/functional/server_tests.js#L978) pass.

mbroadst: I wonder about code like [this](https://github.com/mongodb-js/mongodb-core/blob/master/lib/connection/connect.js#L105-L116), and maybe that we're only checking one path for compression maybe? That definitely looks to me like we're looking for `options.compression` not `options.compressors`

kvwalker: Yeah, that code looks like it should break. I guess we're not testing that anywhere?

I can support both `compression` and `compressors` and file a ticket to move things from checking/using `compression` to `compressors`?

mbroadst: iirc, `compression` was added to wrap `compressors`, and `zlibCompressionLevel`. I'm not sure whether or not we should remove it at this point, but it's probably still possible to pass the uri spec tests using `compression`. It would be great if you could actually write a test that covers what we found here today though.

lib/uri_parser.js
mbroadst: @daprahamian can we resolve this?

test/tests/unit/connection_string_tests.js
kvwalker: I added a test in the [native PR](https://github.com/mongodb/node-mongodb-native/pull/1929).

PR #411, opened 2019-02-22
Pin Servers during Sharded Transactions

daprahamian:
Changes to ensure that all commands in a sharded transaction run against a single `mongos`.

Items to still do:

+ [ ] Ensure that pins are released at the appropriate times
+ [ ] Add support for new SDAM layer

### DIFF
--- a/lib/sdam/topology.js
+++ b/lib/sdam/topology.js
@@ -206,7 +206,7 @@ class Topology extends EventEmitter {
     translateReadPreference(options);
     const readPreference = options.readPreference || ReadPreference.primary;
 
-    this.selectServer(readPreferenceServerSelector(readPreference), (err, server) => {
+    this.selectServer(readPreferenceServerSelector(readPreference), options, (err, server) => {
       if (err) {
         if (typeof callback === 'function') {
           callback(err, null);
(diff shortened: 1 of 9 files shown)


lib/sdam/topology.js
```diff
@@ -310,14 +310,29 @@ class Topology extends EventEmitter {
       options
     );
 
+    const isSharded = this.description.type === TopologyType.Sharded;
+    const session = options.session;
+    const transaction = session && session.transaction;
+
+    if (isSharded && transaction && transaction.server) {
```
daprahamian: `mongos.js` has a protection here to ensure that the pinned server is actually connected. Should we do that here?

lib/transactions.js
```diff
@@ -122,13 +138,30 @@ class Transaction {
...
+      }
       return;
     }
 
     throw new MongoError(
       `Attempted illegal state transition from [${this.state}] to [${nextState}]`
     );
   }
+
+  get isPinned() {
```
daprahamian: it looks like `get isPinned` is defined twice here, and not actually used.

lib/sdam/topology.js
mbroadst: I think that's because the old mongos doesn't have proper server selection. Anyway, I think it's too defensive to check here and I'd rather fix this holistically in our new topology if the issue comes up.

lib/transactions.js
mbroadst: whoops, removing

PR #415, opened 2019-03-07
chore: disable tls renegotiation where possible

daprahamian:
Fixes NODE-1841.

I currently have no way to test this in Core, hence this being WIP.

### DIFF
--- a/lib/connection/connect.js
+++ b/lib/connection/connect.js
@@ -236,6 +236,9 @@ function makeConnection(family, options, callback) {
       socket = tls.connect(parseSslOptions(family, options));
     } else {
       socket = net.createConnection(parseConnectOptions(family, options));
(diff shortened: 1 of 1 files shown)


lib/connection/connect.js
```diff
@@ -236,6 +236,9 @@ function makeConnection(family, options, callback) {
       socket = tls.connect(parseSslOptions(family, options));
     } else {
       socket = net.createConnection(parseConnectOptions(family, options));
+      if (typeof socket.disableRenegotiation === 'function') {
```
mbroadst: aha, related to our previous discussion, you are calling this on the _non_ TLS socket :) It might make sense to move this outside of the `useSsl` block, but I think it only applies to tls sockets.

PR #416, opened 2019-03-08
refactor(pool): remove all sessions-related methods out of pool

mbroadst:
NODE-1892


lib/sessions.js
```diff
@@ -595,9 +599,94 @@ class ServerSessionPool {
...
+    command.txnNumber = BSON.Long.fromNumber(serverSession.txnNumber);
+  }
+
+  // now attempt to apply transaction-specific sessions data
+  if (!inTransaction) {
+    if (session.transaction.state !== TxnState.NO_TRANSACTION) {
+      session.transaction.transition(TxnState.NO_TRANSACTION);
+    }
+
+    // for causal consistency
```
daprahamian: Nit: Add a note here that we should not be doing this for write commands

mbroadst: done

PR #417, opened 2019-03-11
NODE-1828: support passing callbacks to topology destroyed methods

mbroadst:
This is the "core" component of NODE-1828, allowing support for a callback passed to a topology (and subsequently pool and connection) so that we can synchronize resources on client close.

### DIFF
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -193,21 +193,34 @@ class Connection extends EventEmitter {
    * Destroy connection
    * @method
    */
-  destroy() {
+  destroy(options, callback) {
+    if (typeof options === 'function') {
+      callback = options;
+      options = {};
+    }
+
+    options = Object.assign({ force: false }, options);
+
     if (connectionAccounting) {
       deleteConnection(this.id);
     }
 
-    if (this.socket) {
-      // Catch posssible exception thrown by node 0.10.x
-      try {
-        this.socket.end();
-      } catch (err) {} // eslint-disable-line
-
-      this.socket.destroy();
+    if (this.socket == null) {
+      this.destroyed = true;
+      return;
     }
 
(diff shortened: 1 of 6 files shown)


lib/connection/connection.js
```diff
@@ -193,21 +193,34 @@ class Connection extends EventEmitter {
    * Destroy connection
    * @method
    */
-  destroy() {
+  destroy(options, callback) {
```
daprahamian: Is a synchronous error possible at all here? If so, we should catch it and either pass it to the callback or rethrow.

lib/connection/pool.js
```diff
@@ -635,34 +635,53 @@ Pool.prototype.unref = function() {
...
-  });
 
-  // Zero out all connections
-  self.inUseConnections = [];
-  self.availableConnections = [];
-  self.connectingConnections = 0;
-
-  // Set state to destroyed
-  stateTransition(self, DESTROYED);
+    conn.destroy(options, connectionDestroyed);
```
daprahamian: Is it possible for `pool.destroy()` to be called synchronously anymore?

lib/connection/connection.js
mbroadst: by a call to `this.socket.destroy()` you mean? I don't think so (don't see anything in the docs for it).

lib/connection/pool.js
mbroadst: I think the legacy behavior is maintained in the event a callback is not provided. The only behavioral change there is that previously the cleanup would happen most likely before the individual connections were cleaned up (though they would be prevented from being garbage collected because of their references here). Are you worried about something in particular?

daprahamian: I was mostly concerned about the legacy behavior.

Issue #213, opened 2017-08-16
Can round-robin connection pool logic be smarter?


bompus:
Currently, as I read it and have tested, it appears that the pool logic is doing a simple round robin approach. This works well if you don't have a mix of slow and fast queries.

When you have some slow queries, you can easily end up with many fast queries stuck behind a connection that is processing a slow query.

It appears there is already support for the pool to grow to a maximum size of connections, but even using a large connection pool max size doesn't solve the problem. If you have hundreds or thousands of queries or commands executing, at some point, you will still get backed up unnecessarily behind a slow query, regardless of whether you have a max pool size of 10 or 100.

Would it be possible to refactor the pool mechanics to queue up items to the connection with the lowest amount of commands queued up?

You'd probably have to maintain a count of pending commands for each connection, but I believe the benefit would be substantial for anybody who is using the driver with a connection pool > 1 and has a mix of slow and fast queries.

Thank you for considering my suggestion!

mbroadst:
@bompus sounds like a very interesting idea! Unfortunately we are in a push right now to get 3.6 features out the door, so I don't suspect there will be any active work on this, but I would love to collaborate and would definitely be willing to review any contributions you might have in the meantime.

bompus:
@mbroadst It's been a while, but I think this is still relevant. Submitted a PR.

Ijmir:
+1

mbroadst:
Hey @bompus and @Ijmir, we've completely redesigned the connection pool and will be merging it in shortly (see this PR for the initial steps: https://github.com/mongodb/node-mongodb-native/pull/2047). This project has been merged into the main driver, and will soon be outright deprecated. I'm disabling GitHub issues within the week, so please reach out to us through JIRA for further discussion if the new pool design doesn't address your needs.

PR #419, opened 2019-03-13
Adds load balancing to connection pool

bompus:
Related to #213 

Load balance our pool connections, sending next workItem to connection with least number of already-pending workItems. Eliminates need for connectionIndex.

### DIFF
--- a/lib/connection/pool.js
+++ b/lib/connection/pool.js
@@ -141,8 +141,6 @@ var Pool = function(topology, options) {
 
   // Number of consecutive timeouts caught
   this.numberOfConsecutiveTimeouts = 0;
-  // Current pool Index
-  this.connectionIndex = 0;
 
   // event handlers
   const pool = this;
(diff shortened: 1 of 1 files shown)


owenallenaz:
I totally support this. A long, long time ago, I remember we tried to alter the pool so that instead of pushing to connection, we queued the tasks up into an array, and then when a query finished, the worker would pull the next work job off the queue, that way it wouldn't matter if you had slow or fast items in the work queue, since as soon as a worker finished, it would grab the next job available. When we proposed that change a while ago it was just before they did their big rewrite a year or two ago to the pool and it kinda disappeared since they weren't interested in merging a change like that into a dying code pathway. Is a queue like that at all possible with their current architecture or is it just too technically complicated. If it's too tricky, then this is a great interim solution.

bompus:
@owenallenaz They are planning on rewriting the pool logic to match a unified spec across drivers. If you read https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#waitqueue , it seems to be very close, if not identical, to what you described as the ideal scenario, with connections pulling an item from a single WaitQueue as they become available. If that is indeed the case, this PR serves no purpose after a single WaitQueue is implemented as I read the spec.

PR #424, opened 2019-04-07
APM for "serialize command" step

dschenkelman:
Opening up this PR for discussion, not meant to be the final version.

# What?
This changes creates three new APM events, related to the `serializeCommand` operation.

# Why?
The `CommandXEvent`s are very good for high level visibility but lack details. As an engineer that has developed with node.js + MongoDB for a while, I've usually found myself wanting to separate:

1. MongoDB server/DB side command execution time
1. MongoDB client side command execution time delays, especially related to serialization

Item #2 can be particularly useful to identify either updates/inserts of large document sets (which are bad because serialization time blocks the event loop) and large result sets from queries (not in this PR).

# Next steps
Would love to get the team's feedback to learn:

- If this is believed to be useful
- If this is the best approach to achieve this. For example we could alternatively move all the logic into `serializeCommand`, which would be appropriate if other calls are added in the future.

I already have the code ready to relay the events in the (native)[https://github.com/mongodb/node-mongodb-native] repo as well. This is what the events look like when logged to the console:
```
serializationSucceeded SerializeCommandSucceededEvent {
  duration: 7.135826,
  commandName: 'insert',
  buffersLength: 2097337,
  requestId: 2,
  connectionId: 'localhost:27017' }
```

In this case as a developer I could consider the `duration` and `buffersLength` to be fairly high. This can then be either sent to metrics, logs and even tracing services with the appropriate context tracking.

Also, if this is useful I'd like to explore doing the same thing for "deserialization time", aiming to expose timing and size metrics about query results, e.g. knowing about `find` queries that returns > 1MB of size and block the event loop while deserializing such a big chunk of data.

### DIFF
--- a/lib/connection/apm.js
+++ b/lib/connection/apm.js
@@ -190,17 +190,21 @@ class CommandSucceededEvent {
    * @param {Object} command the command
    * @param {Object} reply the reply for this command from the server
    * @param {Array} started a high resolution tuple timestamp of when the command was first sent, to calculate duration
+   * @param {Array} metadata an array of command metadata
    */
-  constructor(pool, command, reply, started) {
+  constructor(pool, command, reply, started, metadata) {
     const cmd = extractCommand(command);
     const commandName = extractCommandName(cmd);
+    metadata = Array.isArray(metadata) ? metadata : [];
+    const commandMetadata = new CommandMetadata(metadata);
 
     Object.assign(this, {
       duration: calculateDurationInMs(started),
       commandName,
       reply: maybeRedact(commandName, extractReply(command, reply)),
       requestId: command.requestId,
-      connectionId: generateConnectionId(pool)
+      connectionId: generateConnectionId(pool),
+      metadata: commandMetadata
     });
   }
 }
@@ -214,23 +218,60 @@ class CommandFailedEvent {
    * @param {Object} command the command
    * @param {MongoError|Object} error the generated error or a server error response
    * @param {Array} started a high resolution tuple timestamp of when the command was first sent, to calculate duration
+   * @param {Array} metadata an array of command metadata
    */
-  constructor(pool, command, error, started) {
+  constructor(pool, command, error, started, metadata) {
     const cmd = extractCommand(command);
     const commandName = extractCommandName(cmd);
+    metadata = (Array.isArray(metadata) && metadata) || [];
+    const commandMetadata = new CommandMetadata(metadata);
 
     Object.assign(this, {
       duration: calculateDurationInMs(started),
       commandName,
       failure: maybeRedact(commandName, error),
       requestId: command.requestId,
-      connectionId: generateConnectionId(pool)
+      connectionId: generateConnectionId(pool),
+      metadata: commandMetadata
     });
+  }
+}
+
+class BSONSerializationData {
+  /**
+   * Create serialize operation data, to store data about the serialize operation within a command's execution
+   *
+   * @param {Array} serializedBuffers the array of buffers that were serialized
+   * @param {Array} started a high resolution tuple timestamp of when the command started to serialize, to calculate duration
+   */
+  constructor(serializedBuffers, started) {
+    serializedBuffers = serializedBuffers || [];
+    started = started || [];
+    Object.assign(this, {
+      type: 'bsonSerialization',
+      buffersLength: serializedBuffers.reduce((s, b) => s + b.length, 0),
+      duration: calculateDurationInMs(started)
+    });
+  }
+}
+
+class CommandMetadata {
+  /**
+   * Create command metadata, to store data about operations that take place within a command's execution
+   *
+   * @param {Array} stepsData an array of command step data
+   */
+  constructor(stepsData) {
+    stepsData.reduce((self, step) => {
+      self[step.type] = step;
+      return self;
+    }, this);
   }
 }
 
 module.exports = {
   CommandStartedEvent,
   CommandSucceededEvent,
-  CommandFailedEvent
+  CommandFailedEvent,
+  BSONSerializationData
 };
--- a/lib/connection/pool.js
+++ b/lib/connection/pool.js
@@ -895,6 +895,8 @@ Pool.prototype.write = function(command, options, cb) {
   // Get the requestId
   operation.requestId = command.requestId;
 
+  const metadata = [];
+  let serializeStarted;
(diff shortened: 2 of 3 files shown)


mbroadst:
Hi @dschenkelman! Cool idea. How would you feel about including this information as metadata on the `Command*Event`s, rather than creating new event types?

dschenkelman:
> Hi @dschenkelman! Cool idea. How would you feel about including this information as metadata on the `Command*Event`s, rather than creating new event types?

@mbroadst really like that idea, thank you for the feedback. Will implement the changes necessary when I get some time 😄

dschenkelman:
Would love your feedback with on the new approach @mbroadst

lib/connection/apm.js
```diff
@@ -190,8 +190,9 @@ class CommandSucceededEvent {
    * @param {Object} command the command
    * @param {Object} reply the reply for this command from the server
    * @param {Array} started a high resolution tuple timestamp of when the command was first sent, to calculate duration
+   * @param {Array} stepsData an array of command step data
    */
-  constructor(pool, command, reply, started) {
+  constructor(pool, command, reply, started, ...stepsData) {
```
daprahamian: We need to support Node 4, so we cannot use the `...` operator

lib/connection/apm.js
```diff
@@ -224,13 +227,45 @@ class CommandFailedEvent {
...
+}
+
+class SerializeStepData {
+  /**
+   * Create serialize operation data, to store data about the serialize operation within a command's execution
+   *
+   * @param {Array} serializedBuffers the array of buffers that were serialized
+   * @param {Array} started a high resolution tuple timestamp of when the command started to serialize, to calculate duration
+   */
+  constructor(serializedBuffers = [], started = []) {
```
daprahamian: also b/c of node 4, no default constructor values.

lib/connection/apm.js
```diff
@@ -224,13 +227,45 @@ class CommandFailedEvent {
...
       failure: maybeRedact(commandName, error),
       requestId: command.requestId,
-      connectionId: generateConnectionId(pool)
+      connectionId: generateConnectionId(pool),
+      stepsData: new CommandStepsData(stepsData)
+    });
+  }
+}
+
+class SerializeStepData {
```
daprahamian: I'd rename this to `BSONSerializationData`.

lib/connection/apm.js
```diff
@@ -224,13 +227,45 @@ class CommandFailedEvent {
...
+  constructor(serializedBuffers = [], started = []) {
+    Object.assign(this, {
+      type: 'serialize',
+      buffersLength: serializedBuffers.reduce((s, b) => s + b.length, 0),
+      duration: calculateDurationInMs(started)
     });
   }
 }
 
+class CommandStepsData {
```
daprahamian: I'd rename this to `CommandMetaData`

lib/connection/apm.js
```diff
@@ -200,7 +201,8 @@ class CommandSucceededEvent {
       commandName,
       reply: maybeRedact(commandName, extractReply(command, reply)),
       requestId: command.requestId,
-      connectionId: generateConnectionId(pool)
+      connectionId: generateConnectionId(pool),
+      stepsData: new CommandStepsData(stepsData)
```
daprahamian: I'd just name the field `metadata` here, so we could use it for other metadata items.

dschenkelman:
> Hi @dschenkelman,
> 
> This is a pretty awesome approach. A few comments

Thanks for the feedback. Will work on it. Sorry for not replying before, was on vacation, etc. 😄

lib/connection/apm.js
```diff
@@ -190,17 +190,21 @@ class CommandSucceededEvent {
    * @param {Object} command the command
    * @param {Object} reply the reply for this command from the server
    * @param {Array} started a high resolution tuple timestamp of when the command was first sent, to calculate duration
+   * @param {Array} metadata an array of command metadata
    */
-  constructor(pool, command, reply, started) {
+  constructor(pool, command, reply, started, metadata) {
     const cmd = extractCommand(command);
     const commandName = extractCommandName(cmd);
+    metadata = (Array.isArray(metadata) && metadata) || [];
```
dschenkelman: @daprahamian how do you handle these internally? the alternative is to throw if a bad type is passed as metadata.

lib/connection/apm.js
```diff
@@ -214,23 +218,60 @@ class CommandFailedEvent {
...
 
+class CommandMetadata {
+  /**
+   * Create command metadata, to store data about operations that take place within a command's execution
+   *
+   * @param {Array} stepsData an array of command step data
+   */
+  constructor(stepsData) {
+    stepsData.reduce((self, step) => {
+      self[step.type] = step;
```
dschenkelman: would you also `delete step.type`? otherwise you end up with some duplication:
``` javascript
metadata: { 'bsonSerialization' : { type: 'bsonSerialization', ... } }
```

dschenkelman:
@daprahamian this is what the structure looks like now:
``` javascript
succeeded CommandSucceededEvent {
  duration: 1.702143,
  commandName: 'endSessions',
  reply: { ok: 1 },
  requestId: 4,
  connectionId: 'localhost:27017',
  metadata: 
   CommandMetadata {
     bsonSerialization: 
      BSONSerializationData {
        type: 'bsonSerialization',
        buffersLength: 141,
        duration: 0.091494 } } }
```

Left a couple of comments as I am not sure about a couple of things.

lib/connection/apm.js
dschenkelman: updated

daprahamian:
This looks awesome. thanks @dschenkelman! One more complication:

In between when you started this PR and now, we have begun the process of merging mongodb-core into node-mongodb-native. You can see in https://github.com/mongodb/node-mongodb-native/tree/next that the main driver no longer depends on `mongodb-core`. Instead, the core components all live inside `lib/core/`.

I can take care of copying it over if you like, but if I do you will lose credit for the commit. Do you want to make the PR against `node-mongodb-native#next` yourself, or would you like me to take care of it?

LMK what works best.
\- Dan

daprahamian:
Hey, @dschenkelman, wanted to follow up. Do you have thoughts on how we want to merge this?

dschenkelman:
@daprahamian thanks for the ping. https://github.com/mongodb/node-mongodb-native/pull/2037

PR #425, opened 2019-04-12
fix(command_result): return entire result in toJSON

kvwalker:
Fixes NODE-1942

### DIFF
--- a/lib/connection/command_result.js
+++ b/lib/connection/command_result.js
@@ -19,7 +19,9 @@ var CommandResult = function(result, connection, message) {
  * @return {object}
  */
 CommandResult.prototype.toJSON = function() {
-  return this.result;
+  let result = this;
+  delete result.message;
+  return result;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


lib/connection/command_result.js
```diff
@@ -19,7 +19,9 @@ var CommandResult = function(result, connection, message) {
  * @return {object}
  */
 CommandResult.prototype.toJSON = function() {
-  return this.result;
+  let result = this;
```
daprahamian: I think we should be making a copy here. Deleting the message b/c of a toJSON call seems like an unintended side effect.

kvwalker: Good catch, thanks!

kvwalker:
I'm a fan of #2. I think that makes the most sense!

kvwalker:
I added https://github.com/mongodb/node-mongodb-native/pull/1977 to test this.

mbroadst:
@kvwalker I was holding off on LGTM'ing this because the native PR is all red in CI, specifically related to checking the result of an `insertOne`

kvwalker:
The native PR is showing that the way `toJSON` currently works is broken so that PR will be green after this PR is merged.

mbroadst:
😱 well don't I feel foolish..

PR #427, opened 2019-04-23
feat(readPreference): add ReadPreference.fromOptions

kvwalker:
Native PR: https://github.com/mongodb/node-mongodb-native/pull/1985

### DIFF
--- a/lib/topologies/read_preference.js
+++ b/lib/topologies/read_preference.js
@@ -104,6 +104,27 @@ const VALID_MODES = [
 ];
 
 /**
+ * Construct a ReadPreference given an options object.
+ *
+ * @param {object} options The options object from which to extract the read preference.
+ * @return {ReadPreference}
+ */
+ReadPreference.fromOptions = function(options) {
+  let readPreference = options.readPreference;
+  if (typeof readPreference === 'string') {
+    return new ReadPreference(readPreference);
+  } else if (!(readPreference instanceof ReadPreference) && typeof readPreference === 'object') {
+    const mode = readPreference.mode || readPreference.preference;
+    if (mode && typeof mode === 'string') {
+      return new ReadPreference(mode, readPreference.tags, {
+        maxStalenessSeconds: readPreference.maxStalenessSeconds
+      });
+    }
+  }
+  return readPreference;
+};
+
+/**
  * Validate if a mode is legal
  *
  * @method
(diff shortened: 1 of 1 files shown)


lib/topologies/read_preference.js
```diff
@@ -103,6 +103,27 @@ const VALID_MODES = [
...
 ];
 
+/**
+ * Construct a ReadPreference given an options object.
+ *
+ * @param {object} options The options object from which to extract the read preference.
+ * @return {ReadPreference}
+ */
+ReadPreference.fromOptions = function(options) {
+  let readPreference = options.readPreference;
```
mbroadst: I'd prefer to be explicit about when we return `null`:
```
const readPreference = options.readPreference;
if (readPreference == null) {
  return null;
}

if (typeof readPreference === 'string') {
  return new ReadPreference(readPreference);
}

if (...) {

}
```

lib/topologies/read_preference.js
```diff
@@ -103,6 +103,27 @@ const VALID_MODES = [
   null
 ];
 
+/**
+ * Construct a ReadPreference given an options object.
+ *
+ * @param {object} options The options object from which to extract the read preference.
+ * @return {ReadPreference}
+ */
+ReadPreference.fromOptions = function(options) {
```
mbroadst: `readPreferenceTags` are supported in the connection string by themselves, should we accommodate that case here? This might be a question posed to the rest of team as well: what is the default read preference when just the tags are specified (I would guess `primary`?)

kvwalker: I don't think that `readPreferenceTags` actually should be supported by themselves. According to the [URI options spec](https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst#List-of-specified-options#readPreferenceTags), `readPreferenceTags` are "only valid if the read preference mode is not primary". To me that implies that `readPreferenceTags` should not be specified if a `mode` is not also specified. The description of `readPreference` in the [server selection spec](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#read-preference) supports this too.

mbroadst: Okay cool. That still means this method needs to check for a `readPreferenceTags` property in `options` - we should ideally use this method to parse a `ReadPreference` from the options derived from the connection string.

PR #429, opened 2019-04-30
fix(transactions): send recovery token on abortTransaction

daprahamian:
Fixes NODE-1961

### DIFF
--- a/lib/sessions.js
+++ b/lib/sessions.js
@@ -474,7 +474,7 @@ function endTransaction(session, commandName, callback) {
   }
 
   if (
-    commandName === 'commitTransaction' &&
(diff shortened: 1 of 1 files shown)


lib/sessions.js
```diff
@@ -474,7 +474,7 @@ function endTransaction(session, commandName, callback) {
   }
 
   if (
-    commandName === 'commitTransaction' &&
+    // Assumption here that commandName is "commitTransaction" or "abortTransaction"
```
kvwalker: Is there a way to verify this assumption?

daprahamian: the function is currently only called with `'commitTransaction'` and `'abortTransaction'`. So as of right now, we don't need to check. I left the comment in case we ever add a 3rd option.

PR #430, opened 2019-05-02
fix authentication with promoteBuffers: true

sebdeckers:
Fixes an error when connecting to MongoDB Atlas with the option `promoteBuffers: true`. Emabling that option causes the binary payload to be returned as a Node.js Buffer instance.

### DIFF
--- a/lib/auth/scram.js
+++ b/lib/auth/scram.js
@@ -177,7 +177,8 @@ class ScramSHA extends AuthProvider {
         return callback(tmpError, null);
       }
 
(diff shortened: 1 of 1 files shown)


sebdeckers:
I'm not sure what you mean. Does line 202 not already address that?

daprahamian:
ahh, nvm, looks like it is not necessary on 213.

daprahamian:
Thank you for contributing @sebdeckers!

PR #434, opened 2019-05-14
fix(connect): fixed syntax issue in connect error handler

daprahamian:
Connect error handler was attempting to read properties off
of an error that had a chance of being null. Removed that,
and replaced the error generation to be more explicit.

Fixes NODE-1960

### DIFF
--- a/lib/connection/connect.js
+++ b/lib/connection/connect.js
@@ -281,10 +281,9 @@ function makeConnection(family, options, _callback) {
   const errorEvents = ['error', 'close', 'timeout', 'parseError', 'connect'];
   function errorHandler(eventName) {
     return err => {
-      if (err == null || err === false) err = true;
       errorEvents.forEach(event => socket.removeAllListeners(event));
       socket.removeListener('connect', connectHandler);
-      callback(new MongoNetworkError(err.message), eventName);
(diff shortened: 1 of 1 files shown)


lib/connection/connect.js
```diff
@@ -355,4 +354,17 @@ function authenticate(conn, credentials, callback) {
   });
 }
 
+function connectionFailureError(type, err) {
+  switch (type) {
+    case 'error':
+      return new MongoNetworkError(err);
```
mbroadst: The reason that I made this `MongoNetworkError(err.message)` in the first place is because the constructor for `MongoError` (which is what the args are eventually passed down to) will convert this _not_ to a `MongoNetworkError`, but instead something that looks like:
```
MongoError {
   name: 'MongoNetworkError'
   ...
}
```

Can you confirm that you haven't undone that here with a test, and imagine a solution if you verify that it is indeed still an issue?

daprahamian: It passes the test `should emit `MongoNetworkError` for network errors`, and I've confirmed that it is the error generated at line 360 that is being asserted on.

mbroadst: indeed it does! I wonder what issue I was running into..

PR #436, opened 2019-05-22
fix(uri-parser): always treat appname as a string

daprahamian:
Fixes NODE-1983

Needs to be ported to `next`

### DIFF
--- a/lib/uri_parser.js
+++ b/lib/uri_parser.js
@@ -124,6 +124,10 @@ function parseQueryStringItemValue(key, value) {
     // deduplicate and simplify arrays
     value = value.filter((v, idx) => value.indexOf(v) === idx);
     if (value.length === 1) value = value[0];
+  } else if (STRING_OPTIONS.has(key)) {
+    // TODO: refactor function to make this early return not
+    // stand out
+    return value;
   } else if (value.indexOf(':') > 0) {
     value = value.split(',').reduce((result, pair) => {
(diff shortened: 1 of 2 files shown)


lib/uri_parser.js
```diff
@@ -124,6 +124,10 @@ function parseQueryStringItemValue(key, value) {
     // deduplicate and simplify arrays
     value = value.filter((v, idx) => value.indexOf(v) === idx);
     if (value.length === 1) value = value[0];
+  } else if (STRING_OPTIONS.has(key)) {
+    // TODO: refactor function to make this early return not
+    // stand out
+    return value;
   } else if (value.indexOf(':') > 0) {
```
mbroadst: I think we also only want to take this path if the key is `readpreferencetags `

daprahamian: also `authMechanismProperties`

mbroadst: should we special case those two explicitly?

daprahamian: I tried doing it pretty quickly, and it led to errors. I think we should instead make a larger ticket to deal with this, where we hard-code everything from https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst, and specifically parse based off of that. For now, I'd like to just get this in to fix the ticket.

mbroadst: OK SGTM

Issue #440, opened 2019-06-11
When cursorId is already a Long, don't call FromNumber


jasontho-ms:
Commit [14b597bc6604e1b7ec69ae786619e979](https://github.com/mongodb-js/mongodb-core/commit/e7e1775fc5175a331ff06bcba543c09285462ffe#diff-14b597bc6604e1b7ec69ae786619e979) introduced an issue:

When the cursorId is already a large-valued Long, calling fromNumber with it will truncate it. This is tested for in other places but not on the lines added by the commit above.

daprahamian:
Closed by #441

PR #441, opened 2019-06-11
fix(cursor): do not truncate an existing Long

jasontho-ms:
This is a fix for issue #440

### DIFF
--- a/lib/cursor.js
+++ b/lib/cursor.js
@@ -710,7 +710,7 @@ function initializeCursor(cursor, callback) {
 
       // Otherwise fall back to regular find path
       const cursorId = result.cursorId || 0;
-      cursor.cursorState.cursorId = Long.fromNumber(cursorId);
+      cursor.cursorState.cursorId = cursorId instanceof Long ? cursorId : Long.fromNumber(cursorId);
       cursor.cursorState.documents = result.documents;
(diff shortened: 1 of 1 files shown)


daprahamian:
@jasontho-ms are you sure this is necessary? I just ran the following in the terminal and it seems okay.

```js
const bson = require('bson');
const assert = require('assert');

assert.ok(bson.Long.fromNumber(bson.Long.MAX_VALUE).equals(bson.Long.MAX_VALUE))
```

jasontho-ms:
MAX_VALUE has no low bits so it works, but try it with one less than MAX_VALUE so it has low bits:

> bson.Long.MAX_VALUE.toString()
'9223372036854775807'
> const foo = bson.Long.fromString('9223372036854775806')
> foo
Long { _bsontype: 'Long', low_: -2, high_: 2147483647 }
> assert.ok(bson.Long.fromNumber(foo).equals(foo))
AssertionError [ERR_ASSERTION]: false == true
>

daprahamian:
@jasontho-ms Ah, good point.

jasontho-ms:
Internet points don’t matter 😊 I’d appreciate it if you port it over. Thanks for taking a look!

daprahamian:
Merged. Thank you for the contribution @jasontho-ms!

PR #445, opened 2019-07-03
fix(server): bubble up pool 'attemptReconnect' event

vkarpov15:
Re: Automattic/mongoose#7872. Pools emit an 'attemptReconnect' event every time they attempt to reconnect to the server, and [the end user driver is already configured to bubble up the 'attemptReconnect' event](https://github.com/mongodb/node-mongodb-native/blob/3105625c2973210eb531ada35a2ec121d8fe89c2/lib/topologies/server.js), the only thing missing is that the core topology doesn't bubble up the event.

### DIFF
--- a/lib/topologies/server.js
+++ b/lib/topologies/server.js
@@ -480,6 +480,7 @@ Server.prototype.connect = function(options) {
   self.s.pool.on('connect', eventHandler(self, 'connect'));
   self.s.pool.on('reconnect', eventHandler(self, 'reconnect'));
   self.s.pool.on('reconnectFailed', eventHandler(self, 'reconnectFailed'));
+  self.s.pool.on('attemptReconnect', eventHandler(self, 'attemptReconnect'));
 
   // Set up listeners for command monitoring
   relayEvents(self.s.pool, self, ['commandStarted', 'commandSucceeded', 'commandFailed']);


mbroadst:
Hey @vkarpov15, we're trying to move to a world where "connecting" and "reconnecting" are no longer concepts users of this module depend on. Rather, the model will be that an operation is performed, a server is selected for the operation, a connection is checked out of a pool and used. In the event that an operation fails, the process is repeated, but there is never a concept of "reconnecting". The pool will seamlessly ensure in the background that there are adequate connections created, but anything related to connection establishment will be internal to the driver.

This is a model that will be introduced with the updated pool which conforms to our new Connection Management and Pooling spec, so its not in yet, however I want to make sure we aren't adding something to soon-to-be-legacy code that mongoose will depend on. Do you really need to know about a pool reconnect?

vkarpov15:
I don't necessarily need to know, but I can imagine this event being useful. Being able to log what's happening with your database connection under the hood is valuable: easier to debug production issues, easier to explain in the docs, and easier to learn how the driver works.

mbroadst:
Well my point is that, very soon, this will not be how the driver works. There will be a number of events related to connections being created, checked in, checked out, etc (see [here](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#events) for the full list of events). There won't be an event for `reconnecting` because that concept will not exist. I want to make sure we're not adding things for users to depend on here which will break soon thereafter.

vkarpov15:
Ah ok, these events make a lot more sense. Props to whoever put so much thought into them :+1: I'll close this PR

PR #448, opened 2019-07-25
fix(replset): correct legacy max staleness calculation

mbroadst:
We were calculating lastUpdateTime using nanoseconds instead of
milliseconds, despite calculating maxStaleness with a millisecond
lastWriteDate

### DIFF
--- a/lib/topologies/replset.js
+++ b/lib/topologies/replset.js
@@ -414,10 +414,10 @@ var pingServer = function(self, server, cb) {
 
       // Calculate latency
       var latencyMS = new Date().getTime() - start;
+
       // Set the last updatedTime
-      var hrTime = process.hrtime();
-      // Calculate the last update time
-      server.lastUpdateTime = hrTime[0] * 1000 + Math.round(hrTime[1] / 1000);
+      var hrtime = process.hrtime();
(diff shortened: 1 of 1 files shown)


lib/topologies/replset.js
```diff
@@ -414,10 +414,10 @@ var pingServer = function(self, server, cb) {
 
       // Calculate latency
       var latencyMS = new Date().getTime() - start;
+
       // Set the last updatedTime
-      var hrTime = process.hrtime();
-      // Calculate the last update time
-      server.lastUpdateTime = hrTime[0] * 1000 + Math.round(hrTime[1] / 1000);
+      var hrtime = process.hrtime();
+      server.lastUpdateTime = (hrtime[0] * 1e9 + hrtime[1]) / 1e6;
```
daprahamian: nit: constants for `1e9` and `1e6`

mbroadst: honestly in the new code paths everything is well named (e.g. `calculateDurationInMS`), for the legacy code I'd just like to get the fix in

daprahamian: I approved, the nit can be ignored.

